PR #1, opened 2021-12-15
Scrabble Submission - Marian

marianphillips:


src/scrabble.js
```diff
@@ -1,6 +1,98 @@
...
+    if (this.word == null) {
+      return 0
+    }
+
+    const wordTrimmed = this.word.trim()
+
+    return this.scoreLoop(wordTrimmed)
+  }
+
+  scoreDoubleTripleLetter (open, close) {
```
Auenc, line 47: This is excellent! It's great to see you using the arguments to prevent yourself from having duplicated code!

src/scrabble.js
```diff
@@ -1,6 +1,98 @@
...
+    }
+
+    if (this.word.indexOf(open) === -1 && this.word.indexOf(close) === -1) {
+      return 0
+    }
+
+    const letterArray = []
+
+    for (let i = 0; i < this.word.length; i++) {
+      if (this.word[i + 2] !== undefined) {
```
Auenc, line 59: This is an interesting approach. What made you need to check if that position of the word was undefined?

Also, this might be a good usecase of a guard statement.
```js
  if(this.word[i + 2] === undefined) {
     continue // this skips this iteration of the loop
  }
}
```

src/scrabble.js
```diff
@@ -1,6 +1,98 @@
...
+        if (this.word[i] === open && this.word[i + 2] === close) {
+          letterArray.push(this.word[i + 1])
+        }
+      }
+    }
+
+    if (letterArray.length === 0) {
```
Auenc, line 66: Looking at the code below this line, I don't think this is required. If the array is empty, your for-loop wont run.  so the `score` variable will be set to 0 when it is returned.

src/scrabble.js
```diff
@@ -1,6 +1,98 @@
...
+      return 0
+    }
+
+    let score = 0
+
+    if (letterArray.length > 0) {
```
Auenc, line 72: this isn't needed, as the if the length is 0 the for-loop won't run

src/scrabble.js
marianphillips: My thinking was that I wanted to stop the for loop running as it would not be efficient. But I didn't think about it not running when the length was 0 anyway! Good to know!

src/scrabble.js
marianphillips: In another problem, I had to check if array[i + 1] was undefined otherwise the code didn't work. However, here as I am checking if it is equal to a specific value, I realise now it isn't necessary as it is allowed to be undefined!!
I may have been using array[i+1].length which could have caused the problems before??

PR #2, opened 2021-12-15
Omar Submission

alsyoufomar:


src/scrabble.js
```diff
@@ -1,6 +1,55 @@
...
+  J: 8, X: 8, Q: 10, Z: 10, '}': 0, '{': 0, '[': 0, ']': 0
+};
+
 class Scrabble {
-  // Write your implementation here
+  constructor(string) {
+    this.string = string
+  }
+  score() {
+    let sum = 0
```
Auenc, line 12: This is being really nit-picky, but you have a really good solution so I thought I'd mention it. It would be more optimal to have your `let sum = 0` to be underneath your `if (this.string === null)`. That way, the computer doesn't need to create a variable that will never be used (if the string is null)

alsyoufomar: that's true the function will die after it returns something, but I thought it would not affect the function's performance or the memory as the integer takes only 4 bytes, but even though thanks for pointing that.

src/scrabble.js
```diff
@@ -1,6 +1,55 @@
...
+        sum += POINTS[word[n + 1]]
+      }
+      else if (word[n] == '[') {
+        sum += POINTS[word[n + 1]] * 2
+      }
+    }
+    return sum;
+  }
+}
+function double(word) {
```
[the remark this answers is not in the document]
alsyoufomar: I thought of it tho indeed it would have looked more encapsulated (which is better) but in that case, I need to call it a separated method and I was aiming to handle all cases in one call.

PR #3, opened 2021-12-16
scrabble submission - acceptance criteria

usamalodhi:


spec/scrabble-spec.js
```diff
@@ -1,54 +1,114 @@
...
+  });
+
+  it("scores a case-insensitive word", function () {
+    scrabble = new Scrabble("OXYPHENBUTAZONE");
+
+    expect(scrabble.score()).toEqual(41);
+  });
+
+  //Extension criteria Tests
+  //double letter
```
Auenc, line 56: These comments are a little moot. You should ensure your describe/it are descriptive, as that is what will be displayed when the test fails.

src/scrabble.js
```diff
@@ -1,6 +1,111 @@
 class Scrabble {
-  // Write your implementation here
+  constructor (word) {
+    this.word = word
+    this.wordMultiply = word
+    this.scoreMap = {
```
Auenc, line 5: It doesn't look like your `scoreMap` ever changes. This could be placed outside of your constructor/outside of your class.

src/scrabble.js
```diff
@@ -1,6 +1,111 @@
...
+  }
 
+  doubleLetter () {
+    const wordArray = []
+    for (let i = 0; i < this.word.length - 1; i++) {
+      if (this.word.charAt(i) === '{' && this.word.charAt(i + 2) === '}') {
+        wordArray.push(this.word.charAt(i + 1))
+      }
+    }
+    return wordArray.toString() ?? '' // nullish coallescing - return RHS when LHS is falsy
```
Auenc, line 55: This nullish coalescing is not required. When you use `toString()` on an empty array it will return `''`

src/scrabble.js
```diff
@@ -1,6 +1,111 @@
...
+  }
+
+  tripleLetter () {
+    const wordArray = []
+    for (let i = 0; i < this.word.length - 1; i++) {
+      if (this.word.charAt(i) === '[' && this.word.charAt(i + 2) === ']') {
+        wordArray.push(this.word.charAt(i + 1))
+      }
+    }
+    return wordArray.toString() ?? ''
```
Auenc, line 65: same as above

src/scrabble.js
```diff
@@ -1,6 +1,111 @@
...
+  }
+
+  lettersScore () {
+    this.word = this.word // original word
+      .concat(this.doubleLetter()) // + one more letter from orignal word (double letter)
+      .concat(this.tripleLetter())
+      .concat(this.tripleLetter()) // + two more letters from original word (triple letter)
```
Auenc, line 72: I think it would have been a better approach to have your `tripleLetter` method to add the letters twice instead of calling it twice. It looks odd that you're having to call the method twice. However, if you wanted to keep this approach, you could have refactored your code to have a method to add the letters using the symbols as input. Like so:

```js
addLetters(leftSymbol, rightSymbol) {
const wordArray = []
    for (let i = 0; i < this.word.length - 1; i++) {
      if (this.word.charAt(i) === leftSymbol && this.word.charAt(i + 2) === rightSymbol) {
        wordArray.push(this.word.charAt(i + 1))
      }
    }
    return wordArray.toString()
}
```

Then you could change your `doubleLetters` and `tripleLetters` methods to use this like so:
```js
doubleLetters() {
  addLetters('{', '}')
}

tripleLetters() {
  addLetters('[', ']')
}
```

The reason this might be better is that you have a lot of duplicated code at the moment.

src/scrabble.js
```diff
@@ -1,6 +1,111 @@
...
+    this.word = this.word.toUpperCase()
+    if (this.word.length === 1) return this.scoreMap[this.word]
+    if (this.word.length > 1) {
+      return [...this.word]
+        .map((letter) => this.scoreMap[letter])
```
Auenc, line 77: This is a little hard to read, I would maybe try to separate this out into another function

src/scrabble.js
```diff
@@ -1,6 +1,111 @@
...
+        .reduce((acc, letterNum) => (acc = acc + letterNum))
+    }
+  }
+
+  wordMultiplier () {
```
Auenc, line 82: I think you could split this method into 2 smaller ones.

Auenc (commented):
Excellent submission! I think in some places you could have maybe simplified things. In future it would be good to try and reduce duplicated code, like with your double and triple letters methods :)

PR #4, opened 2021-12-17
Basic layout started

ben-corbin:


src/scrabble.js
```diff
@@ -1,6 +1,58 @@
+const letterScores = {
+  onePoint: ['a', 'A', 'e', 'E', 'i', 'I', 'o', 'O', 'u', 'U', 'l', 'L', 'n', 'N', 'r', 'R', 's', 'S', 't','T'],
```
Auenc, line 2: This is a good approach. However, you could limit the amount of if statements you have in your scoring method if you structured you data a little bit differently. If you instead had an object containing each of the letters associated with their score, you could then just look up the score. Something like this: 

```js
const letterScores = {
A: 1
B: 1
C: 1
// etc
}
```

Then look up the scores using something similar to this:
```js
let score = 0
  score += letterScores[this.word[i]]
}
```

src/scrabble.js
```diff
@@ -1,6 +1,58 @@
...
+  }
+}
+  if (this.word.charAt(0) === '{' && this.word.charAt(this.word.length - 1) === '}') count *= 2 
+  if (this.word.charAt(0) === '[' && this.word.charAt(this.word.length - 1) === ']') count *= 3 
+  return count
+}
 
 }
 
+scrabble = new Scrabble('dog')
```
Auenc, line 54: I think you left this in from when you were debugging something. Just remember it's always a good idea to delete any debug code you no longer need :)

PR #5, opened 2021-12-17
starting scrabble

KinTale:


Auenc (commented):
Excellent solution! Nothing really to point out at all. Although it would have been nice to see you attempt some of the challenge criteria :)

PR #8, opened 2021-12-17
push

PointWolfAnj:


src/scrabble.js
```diff
@@ -1,6 +1,33 @@
+const letterScores = {
```
Auenc, line 1: This is a good approach. However, you could limit the amount of if statements you have in your scoring method if you structured you data a little bit differently. If you instead had an object containing each of the letters associated with their score, you could then just look up the score. Something like this: 

```js
const letterScores = {
A: 1
B: 1
C: 1
// etc
}
```

Then look up the scores using something similar to this:
```js
let score = 0
  score += letterScores[this.word[i]]
}
```

PR #9, opened 2021-12-17
Paul Scrabble Challenge

therearedoors:


src/scrabble.js
```diff
@@ -1,6 +1,133 @@
...
+let tripleLetterRef = []
+let points = 0
+const commonLetters = [['A', 'E', 'I', 'O', 'U', 'L', 'N', 'R', 'S', 'T'], ['D', 'G'], ['B', 'C', 'M', 'P'], ['F', 'H', 'V', 'W', 'Y'], ['K']]
+const rareLetters = [['J', 'X'], ['Q', 'Z']]
+const double = /\{[A-Z{}[\]]*\}/g
+const triple = /\[[A-Z{}[\]]*\]/g
+
+// functions called directly by Scrabble class
+function noInput (type) {
+  return type === null
```
Auenc, line 13: So this is a little bit of a quirk with javascript (and using === which forces types to be the same). `null` is a value that developers set. However, if nothing is set for the word value it wouldn't be `null` instead it would be `undefined`.

```js
let a = null
let b = undefined

a === null // true
b === null //false
b == null // true
```

src/scrabble.js
```diff
@@ -1,6 +1,133 @@
...
+  for (let i = 0; i < 2; i++) { // "rare" letter scores 8 and 10
+    if (rareLetters[i].includes(element)) {
+      scrabbleLetter(element, (i + 4) * 2)
+    }
+  }
+}
+// function called by scoreCommonLetter and scoreRareLetter
+function scrabbleLetter (element, pts) {
+  points += checkDouble(`{${element}}`, doubleLetterRef) // checks whether double or triple or neither and updates points accordingly
+    ? pts * 2
```
Auenc, line 104: Try to avoid using ternaries. _Especially_ nested ternaries like this. It's very hard to read, and even harder to debug!

src/scrabble.js
```diff
@@ -1,6 +1,133 @@
...
+function scoreCommonLetter (element) { // "common" letter scores from 1-5
+  for (let i = 0; i < 5; i++) {
+    if (commonLetters[i].includes(element)) {
+      scrabbleLetter(element, i + 1)
+    }
+  }
+}
+
+function scoreRareLetter (element) {
+  for (let i = 0; i < 2; i++) { // "rare" letter scores 8 and 10
```
Auenc, line 95: This 2 here is what we call a magic number. Typically it's bad to have them. Obviously there are times where we _need_ to have a hard coded number like this, but it is better to put that value in a constant variable and use the variable. The reason for this is it's easier then to see what this `2` means. Also, if multiple things rely on that magic number, then it's easier to change the value throughout the code if you have a constant variable holding it.

src/scrabble.js
therearedoors: I think I chose to use the nested ternary here because I was trying to meet the '5 line per function' criterion...
I've changed it to this, which is wordier but I guess clearer:
```js
  scrabbleLetter (element, pts) {
    if (this.doubleLetters.includes(`{${element}}`)) {
      this.points += pts * 2
    } 
   else if (this.tripleLetters.includes(`[${element}]`)) {
      this.points += pts * 3
    }
    else this.points += pts
  }
  ```

PR #10, opened 2021-12-18
Will's scrabble challenge

Willcowen:


src/scrabble.js
```diff
@@ -1,6 +1,49 @@
...
-  // Write your implementation here
+  constructor(word) {
+    this.word = word
+  }
+  score () {
+    let count = 0
+    if (this.word === null || this.word.length === 0 || this.word === " \t\n") {
+      return count
+    }
+    for (let i = 0; i < this.word.length; i++) {
```
phillipdunn, line 20: in here you could then handle lower case words like:
`let word = this.word.trim().toUpperCase();`

PR #11, opened 2021-12-20
core criteria scrabble

Khateeb1111:


src/scrabble.js
```diff
@@ -1,5 +1,33 @@
...
+  } else if (this.word === 'f') {
+    return 4
+  } else if (this.word === 'street') {
+    return 6
+  } else if (this.word === 'quirky') {
+    return 22
+  } else if (this.word === 'OXYPHENBUTAZONE') {
+    return 41
+  }
+
```
phillipdunn, line 26: This doesn't look complete?

src/scrabble.js
```diff
@@ -1,5 +1,33 @@
 class Scrabble {
   // Write your implementation here
+constructor(word) {
+this.word = word
+}
+
+
```
phillipdunn, line 7: Maybe you want to add an object in here to be able to iterate through it to get the scores for each letter

PR #12, opened 2021-12-20
Dan - Scrabble-challenge

dan1212111:


src/scrabble.js
```diff
@@ -1,6 +1,48 @@
+const val1 = ["A", "E","I","O","U","L","N","R","S","T","a","e","i","o","u","l","n","r","s","t",];
+const val2 = ["D", "d", "G", "g"];
+const val3 = ["B", "b", "C", "c", "M", "m", "P", "p"];
+const val4 = ["F", "f", "H", "h", "V", "v", "W", "w", "Y", "y"];
+const val5 = ["K", "k"];
+const val8 = ["J", "j", "X", "x"];
+const val10 = ["Q", "q", "Z", "z"];
+
```
phillipdunn, line 8: A better way to do this, which would make it more maintainable, would be to put both the letters and corresponding values as key-value pairs in an object, e.g.
const scoresObject = { A: 1, E: 1, ...etc }

src/scrabble.js
```diff
@@ -1,6 +1,48 @@
...
   // Write your implementation here
 
+  constructor(word) {
+    this.word = word;
+   }
+
+  score() {
+    let score = 0
+
+    if (this.word === undefined || this.word === null) {
```
phillipdunn, line 19: probably also want to handle spaces, like tab or new line space e.g.

src/scrabble.js
```diff
@@ -1,6 +1,48 @@
...
+      return 0
+    }
```
phillipdunn, line 21: in here you could then handle lower case words like:
let word = this.word.trim().toUpperCase();

PR #15, opened 2021-12-21
core-criteria

elvisonob:


src/scrabble.js
```diff
@@ -1,6 +1,50 @@
...
+    lettersF: ['J', 'X', 'j', 'x'],
+    lettersG: ['Q', 'Z', 'q', 'z']
+}
+
 class Scrabble {
-  // Write your implementation here
+    // Write your implementation here
+    constructor(word) {
+        this.word = word
+    }
```
phillipdunn, line 16: in here you could then handle lower case words like:
`let word = this.word.trim().toUpperCase();`

PR #16, opened 2021-12-21
Safia Ibrahim 

New-Arday:


src/scrabble.js
```diff
@@ -1,6 +1,75 @@
...
+  "W",
+  "Y",
+  "K",
+  "J",
+  "X",
+  "Q",
+  "Z",
+];
+const valueArray = [
+  1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 3, 3, 3, 3, 4, 4, 4, 4, 4, 5, 8, 8, 10,
```
phillipdunn, line 30: A better way to do this, which would make it more maintainable, would be to put both the letters and corresponding values as key value pairs in an object, e.g. 
`const scoresObject = {
 A: 1,
  E: 1,
  ...etc
}`

src/scrabble.js
```diff
@@ -1,6 +1,75 @@
...
+];
 class Scrabble {
   // Write your implementation here
 
+  constructor(word) {
+    this.word = word;
+  }
+
+  score() {
+    // this.word
```
phillipdunn, line 41: good sudo coding, really helps understand your thinking

PR #17, opened 2022-01-05
Javascript Scrabble Challenge

JasonKundi:


src/scrabble.js
```diff
@@ -1,6 +1,57 @@
...
+  constructor (word, score) {
+    this.word = word
+  }
+
+  score () {
+    let score = 0
+    if (this.word === null || this.word === '\t\n') {
+      return score
+    }
+    const word = this.word.trim().toUpperCase()
```
Auenc, line 49: if you did this before your if statement - you wouldn't need to check `this.word === '\t\n'`. This is because `trim()` will remove them. Great use of trim by the way!

Auenc (commented):
Excellent solution! Nothing really to add other than my comment about using trim earlier on in the method :)

PR #19, opened 2022-03-11
Gui Rodrigues - Scrabble Module 1

qruz-hq:


src/scrabble.js
```diff
@@ -1,6 +1,55 @@
...
+  score () {
+    let s = 0
+    let doubleNext = false
+    let tripleNext = false
+    if (!this.isValid()) return s
+    this.word
+      .replace(' ', '')
+      .toLowerCase()
+      .split('')
+      .forEach((ch, index) => {
```
Anicaaa, line 27: Interesting. I wonder what 'ch' does. Never used or seen it before.

src/scrabble.js
```diff
@@ -47,8 +40,14 @@ class Scrabble {
         } else s += this.getCharScore(ch)
       })
 
-    if (this.word.indexOf('[') === 0 && this.word.indexOf(']') === this.word.length - 1) return s * 3
-    if (this.word.indexOf('{') === 0 && this.word.indexOf('}') === this.word.length - 1) return s * 2
+    if (
+      this.word.indexOf('[') === 0 &&
+      this.word.indexOf(']') === this.word.length - 1
```
Hoolio333, line 45: Why word length - 1?

src/scrabble.js
```diff
@@ -1,6 +1,35 @@
...
+    let s = 0
+    if (!this.isValid()) return s
+    this.word
+      .replace(' ', '')
+      .toLowerCase()
+      .split('')
+      .forEach((ch) => (s += this.getCharScore(ch)))
+
+    return s
+  }
```
Hoolio333, line 52: I like this layout, very clean & clear - easy to read

src/scrabble.js
```diff
@@ -1,6 +1,35 @@
+const ScoreUtil = require('./utils/Scores')
+
```
Hoolio333, line 2: Not many pushes (I did the same), think we should probably push at every stage more.

src/utils/Scores.js
```diff
@@ -0,0 +1,32 @@
...
+  q: 10,
+  r: 1,
+  s: 1,
+  t: 1,
+  u: 1,
+  v: 4,
+  x: 8,
+  y: 4,
+  w: 4,
+  z: 10
```
Anicaaa, line 27: Good to know there are other ways to code in Javascript. Here, I personally used the For Loop.

src/utils/Scores.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+module.exports = {
+  scores: scores
+}
```
Tibor22, line 32: A good way to keep it modular

src/scrabble.js
Creatorrr1: I think that's for last index

src/scrabble.js
```diff
@@ -12,8 +12,7 @@ class Scrabble {
   }
 
   getCharScore (char) {
-    if (ScoreUtil.scores[char]) return ScoreUtil.scores[char]
-    else return 0
+    return ScoreUtil.scores[char] || 0
```
Creatorrr1, line 15: The or 0 was very efficient and line efficient, it even executed one of the test. Simple and clean code is the way 👌

PR #20, opened 2022-03-11
Arisa Sigrist

sigristarisa:


src/scrabble.js
```diff
@@ -1,6 +1,49 @@
...
+          if (this.fivePoints.includes(wordArr[i])) subScore = 5
+          if (this.eightPoints.includes(wordArr[i])) subScore = 8
+          if (this.tenPoints.includes(wordArr[i])) subScore = 10
+          finalScore += subScore * multiply
+        }
+      }
+    }
+    return finalScore
+  }
 }
```
0xtobycarlo, line 48: no comments or improvements that I can think of, all looks perfect 🙂

src/scrabble.js
```diff
@@ -1,6 +1,49 @@
...
+    this.fourPoints = ['f', 'h', 'v', 'w', 'y']
+    this.fivePoints = ['k']
+    this.eightPoints = ['j', 'x']
+    this.tenPoints = ['q', 'z']
+  }
 
+  score () {
+    let finalScore = 0
+    let subScore = 0
+    let multiply = 1
```
Alex90Jennings, line 16: nice use of a multiple to adjust the scoring system, I think we will implement this a lot when making minigames

src/scrabble.js
```diff
@@ -19,10 +19,9 @@ class Scrabble {
       return finalScore
     } else {
```
AlexDjangoX, line 20: Very elegant solution. clean. Concise.

src/scrabble.js
```diff
@@ -1,6 +1,49 @@
...
+        // check if there is curly bracket or square bracket
+        // If so, change the multiply to 2 or 3. If not, multiply remains 1
+        if (wordArr[i] === '{') {
+          multiply = 2
+        } else if (wordArr[i] === '}') {
+          multiply = 1
+        } else if (wordArr[i] === '[') {
+          multiply = 3
+        } else if (wordArr[i] === ']') {
+          multiply = 1
```
Alex90Jennings, line 32: what would happen if there is a double letter within a double scored word? The multiple would return to one after the double letter

if (wordArr[i] === '{') {
          multiply = multiply * 2
        } else if (wordArr[i] === '}') {
          multiply = multiply / 2
        } else if (wordArr[i] === '[') {
          multiply = multiply * 3
        } else if (wordArr[i] === ']') {
          multiply = multiple / 3

src/scrabble.js
```diff
@@ -1,6 +1,49 @@
...
+          if (this.threePoints.includes(wordArr[i])) subScore = 3
+          if (this.fourPoints.includes(wordArr[i])) subScore = 4
+          if (this.fivePoints.includes(wordArr[i])) subScore = 5
+          if (this.eightPoints.includes(wordArr[i])) subScore = 8
+          if (this.tenPoints.includes(wordArr[i])) subScore = 10
+          finalScore += subScore * multiply
+        }
+      }
+    }
+    return finalScore
```
Alex90Jennings, line 45: very nice and readable solution

PR #22, opened 2022-03-11
Alex J - Scrabble

Alex90Jennings:


src/scrabble.js
```diff
@@ -1,6 +1,22 @@
...
+      const letter = letterValue[this.word[i].toLowerCase()] || 0
+      if (this.word[i] === '[') { multiple = 2 * multiple }
+      if (this.word[i] === '{') { multiple = 3 * multiple }
+      if (this.word[i] === ']') { multiple = multiple / 2 }
+      if (this.word[i] === '}') { multiple = multiple / 3 }
+      totalScore += (letter * multiple)
+    }
+    return totalScore
+  }
 }
```
sigristarisa, line 21: The code is so condensed! Well done!!!

src/scrabble.js
```diff
@@ -1,6 +1,22 @@
+const letterValue = { a: 1, b: 3, c: 3, d: 2, e: 1, f: 4, g: 2, h: 4, i: 1, j: 8, k: 5, l: 1, m: 3, n: 1, o: 1, p: 3, q: 10, r: 1, s: 1, t: 1, u: 1, v: 4, w: 4, x: 8, y: 4, z: 10 }
```
0xtobycarlo, line 1: Could have separate lines for each point value -> 1, 2, 3, 4, 5, 8, 10 for better readability here - for example: `twoPoints: ["d", "D", "g", "G"],`

src/scrabble.js
```diff
@@ -1,6 +1,22 @@
...
+
 class Scrabble {
-  // Write your implementation here
+  constructor (word) { this.word = word }
```
0xtobycarlo, line 5: well formatted and structured code from line 6 to 20

src/scrabble.js
```diff
@@ -1,6 +1,22 @@
...
 
+  score () {
+    let totalScore = 0
+    let multiple = 1
+    if (this.word === null || this.word === '') { return 0 }
```
sigristarisa, line 9: What if this.word is other data types (e.g NaN, false etc)? It might pass the conditionals...
I would suggest doing if(typeof this.word !== "string")

src/scrabble.js
```diff
@@ -1,6 +1,22 @@
...
+    if (this.word === '\t' || this.word === '\n') { totalScore += 0 }
+    for (let i = 0; i < this.word.length; i++) {
```
sigristarisa, line 11: You're right, there is no need to make the word into an array... thanks for making me realize!

src/scrabble.js
```diff
@@ -1,9 +1,7 @@
 const letterValue = { a: 1, b: 3, c: 3, d: 2, e: 1, f: 4, g: 2, h: 4, i: 1, j: 8, k: 5, l: 1, m: 3, n: 1, o: 1, p: 3, q: 10, r: 1, s: 1, t: 1, u: 1, v: 4, w: 4, x: 8, y: 4, z: 10 }
 
 class Scrabble {
-  constructor (word) {
-    this.word = word
-  }
+  constructor (word) { this.word = word }
 
   score () {
     let totalScore = 0
```
AlexDjangoX, line 7: Nice concise solution. Clean code.

PR #24, opened 2022-03-11
SeanRich Scrabble Challenge

SeanRichhh:


src/scrabble.js
```diff
@@ -1,6 +1,55 @@
+const letters = {
+  value1: ["A", "E", "I", "O", "U", "L", "N", "R", "S", "T" ,'a', 'e', 'i', 'o', 'u', 'l', 'n', 'r', 's', 't'],
+  value2: ["D", "G", "d", "g"],
+  value3: ["B", "C", "M", "P", "b" ,"c" ,"m", "p"],
+  value4: ["F", "H", "V", "W", "Y", "f", "h", "v", "w", "y"],
+  value5: ["K", "k"],
+  value8: ["J", "X" ,"j" ,"x"],
+  value10: ["Q", "Z", "q", "z"]
+}
```
d-username, line 9: instead of typing in small and capital letters each, I used .toLowerCase() on the letter in the function.

src/scrabble.js
```diff
@@ -1,6 +1,55 @@
...
+        result = result + 4
+      }
+      if(letters.value5.includes(this.word[i])){
+        result = result + 5
+      }
+      if(letters.value8.includes(this.word[i])){
+        result = result + 8
+      }
+      if(letters.value10.includes(this.word[i])){
+        result = result + 10
```
d-username, line 48: I did my code exactly this way first...then when I added the extension criteria it become way too long. 
then I figured there is a much easier way when you don't need to manually type in the score for each letter but a method can find it for you.

src/scrabble.js
```diff
@@ -1,6 +1,55 @@
...
+    if(this.word === null){
+      return 0
+    }
+    if(this.word === " \t\n"){
+      return 0
+    }
+    let result = 0
+    for(let i = 0; i < this.word.length; i++){
+      if(letters.value1.includes(this.word[i])){
+        result = result + 1
```
Dan2024, line 30: you could change `result = result + 1` to `result += 1` and it would produce the same result. but if you find it easier the way you did it no need to change it 👍

src/scrabble.js
```diff
@@ -1,6 +1,55 @@
...
+        result = result + 5
+      }
+      if(letters.value8.includes(this.word[i])){
+        result = result + 8
+      }
+      if(letters.value10.includes(this.word[i])){
+        result = result + 10
+      }
+    }
+    return result
```
PBara7a, line 51: I agree with David here. You implemented the core criteria nicely and the code is easy to follow, but storing the scores associated with the letters in a different way can make your life easier if you decide to expand on this. I used an object for example, like:
`letters = { A: 1, B: 3, C: 3...}`
Then you can add the value by accessing the key, which is a letter, instead of typing the values directly on your code.

PR #26, opened 2022-03-11
Paulo Barata

PBara7a:


spec/scrabble-spec.js
```diff
@@ -51,4 +51,91 @@ describe("Scrabble", function() {
 
     expect(scrabble.score()).toEqual(41)
   })
+
```
Dan2024, line 54: you added lots of extra testing criteria, with some a lot more complicated entries, nice work!

src/scrabble.js
```diff
@@ -1,6 +1,126 @@
...
+  //                       S: 1, T: 1, U: 1,
+  //                       V: 4, W: 4, X: 8,
+  //                       Y: 4, Z: 10 }
+
+  /**
+   * Iterates over every character in a given word and sums the total score.
+   * @returns A score for the given word.
+   */
+  score () {
+    if (!this.isValid(this.word)) return 0
```
Dan2024, line 52: i like the use of this if statement with the isValid function here - makes it easier to understand whats going on 👍

src/scrabble.js
```diff
@@ -1,6 +1,126 @@
...
 
+  /**
+   * Checks if character is a letter from the alphabet.
+   * @param {string} ch - A character from a string.
+   * @returns Boolean
+   */
+  _isALpha (ch) {
+    return ((ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z'))
+  }
 }
```
SeanRichhh, line 124: Clean code very easy to read and understand good job :)

src/scrabble.js
PBara7a: I got the idea to write this extra method from Gui, it helped in breaking down the score method a bit more and improving readability.

PR #27, opened 2022-03-13
David Czuczor

d-username:


src/scrabble-long.js
```diff
@@ -0,0 +1,134 @@
...
+  // first find out if the string is empty string, whitepsace or null, then score 0.
+  score () {
+    if (
+      this.word === '' ||
+      this.word === ' ' ||
+      this.word === ' \t\n' ||
+      this.word === null
+    ) {
+      this.points += 0
+      return 0
```
Dan2024, line 16: do you need the `this.points += 0` here? i would've thought you could just use `return 0'

src/scrabble-long.js
```diff
@@ -0,0 +1,134 @@
...
+        i++
+      }
+
+      if (this.word[i] === ']') {
+        triplePoints = false
+      }
+
+      if (
+        pointOne.includes(this.word[i]) ||
+        pointOne.includes(this.word[i].toLowerCase())
```
Dan2024, line 53: theres quite a lot of repetition of `.toLowerCase()` here, potentially could you convert this.word to lowercase before you enter it into the for loop? e.g `this.word = this.word.toLowerCase();` on line 50 and then take out all the toLowerCase lines?

spec/scrabble-spec.js
```diff
@@ -1,54 +1,94 @@
...
     expect(scrabble.score()).toEqual(22)
   })
 
-  it('scores a case-insensitive word', function() {
+  it('scores a case-insensitive word', function () {
     scrabble = new Scrabble('OXYPHENBUTAZONE')
 
     expect(scrabble.score()).toEqual(41)
   })
+
```
Dan2024, line 54: nice use of adding lots of extra test cases 👍

src/scrabble.js
```diff
@@ -1,6 +1,44 @@
...
+      }
+    }
+  }
 
+  // first i check for empty strings and white spaces.
+  score () {
+    if (
+      this.word === '' ||
+      this.word === ' ' ||
+      this.word === ' \t\n' ||
```
SeanRichhh, line 31: line 31 doesn't need to be in the if statement, you can completely remove it form your code.

src/scrabble-long.js
```diff
@@ -0,0 +1,134 @@
...
+          this.points += 10 * 2
+        } else if (triplePoints === true) {
+          this.points += 10 * 3
+        } else {
+          this.points += 10
+        }
+      }
+    }
+    return this.points
+  }
```
SeanRichhh, line 131: nice code to read and I understand how you did the extension criteria.

src/scrabble-long.js
```diff
@@ -0,0 +1,134 @@
...
+        pointTen.includes(this.word[i]) ||
+        pointTen.includes(this.word[i].toLowerCase())
+      ) {
+        if (doublePoints === true) {
+          this.points += 10 * 2
+        } else if (triplePoints === true) {
+          this.points += 10 * 3
+        } else {
+          this.points += 10
+        }
```
PBara7a, line 127: Similar to what Dan commented above, there is a lot of repetition here checking if doublePoints or triplePoints is true. You could refactor by replacing those two variables by a single one holding a multiplier value and then multiply the points you are adding by the current multiplier.

PR #28, opened 2022-03-13
Anica Rahman - Scrabble Challenge

Anicaaa:


src/scrabble.js
```diff
@@ -1,43 +1,69 @@
...
-}
-  }
+  score() {
+    //Function in the class is a method. This is a method.
+    let score = 0; //Flexible variable
+    if (typeof this.word !== "string") {
+      //If it's NOT a string, return 0. This is a conditional flow with a strict inequality (!==)
+      return 0;
+    } else {
+      const wordArray = this.word.toUpperCase().split(""); // variable with array capitalising the splitted word.
```
Hoolio333, line 16: Why split?  taking the word out of the string?

src/scrabble.js
```diff
@@ -1,6 +1,9 @@
 class Scrabble {
   // Write your implementation here
 
+  // For this challenge use CLASSES, FUNCTIONS, LOOPS and FLOW CONTROL.
 }
 
 module.exports = Scrabble
+
+// ./node_modules/jasmine/bin/jasmine.js
```
Hoolio333: Not many pushes done (I did the same)

src/scrabble.js
```diff
@@ -1,8 +1,42 @@
...
+  } else if (wordArray[i] === "B" || wordArray[i] === "C" || wordArray[i] === "M" || wordArray[i] === "P") {
+  score += 3;
+  } else if (wordArray[i] === "F" || wordArray[i] === "H" || wordArray[i] === "V" || wordArray[i] === "W" || wordArray[i] === "Y" ) {
+  score += 4;
+  } else if (wordArray[i] === "K") {
+  score += 5;
+  } else if (wordArray[i] === "J" || wordArray[i] === "X") {
+  score += 8;
+  } else if (wordArray[i] === "Q" || wordArray[i] === "Z") {
+    score += 10
```
Hoolio333: Looks clear, good use of if & else statements

src/scrabble.js
```diff
@@ -1,8 +1,42 @@
-class Scrabble {
-  // Write your implementation here
+// For this challenge use CLASSES, FUNCTIONS, LOOPS and FLOW CONTROL.
+// We have the word "street" so we need to slipt it down into letters. We want to check the alphabets. So we use .split() in an array.
```
Hoolio333, line 2: Nice use of pseudo code

src/scrabble.js
```diff
@@ -1,43 +1,69 @@
...
+          score += 4;
+        } else if (wordArray[i] === "K") {
+          score += 5;
+        } else if (wordArray[i] === "J" || wordArray[i] === "X") {
+          score += 8;
+        } else if (wordArray[i] === "Q" || wordArray[i] === "Z") {
+          score += 10;
+        }
+      }
+    }
```
Creatorrr1, line 57: Nice way of doing it, it's really an intriguing way. However, in point-system try to use objects and arrays for letter scores rather than using conditional-flow because you would lose modularity and adding extra values would be difficult. Other than that it wasn't deal breaker since we're all learning and finding different ways of doing it will make growth exponentially faster.

src/scrabble.js
```diff
@@ -1,43 +1,69 @@
...
-  score += 4;
-  } else if (wordArray[i] === "K") {
-  score += 5;
-  } else if (wordArray[i] === "J" || wordArray[i] === "X") {
-  score += 8;
-  } else if (wordArray[i] === "Q" || wordArray[i] === "Z") {
-    score += 10
-  }
-}
-  }
```
Creatorrr1: Your such a pro at using if else statement, it's amazing so keep it up.

src/scrabble.js
qruz-hq: It's probably because it's used to iterate through the chars of the string afterwarss

PR #29, opened 2022-03-14
Dan Northcott

Dan2024:


src/scrabble.js
```diff
@@ -1,6 +1,70 @@
...
+      //   const bracketIndexClosed = this.word.indexOf("}");
+      //   this.word.slice(bracketIndexOpen, bracketIndexClosed);
+      //   // if (points10.includes(this.word[bracketIndex + 1])) {
+      //   //   score += 10;
+      //   // }
+      // }
+    }
+    return score;
+  }
 }
```
PBara7a, line 68: Your code looks clean and it is easy to read.

PBara7a, line 68: Your code looks clean and it is easy to read.

src/scrabble.js
```diff
@@ -1,6 +1,70 @@
...
+    for (let i = 0; i < this.word.length; i++) {
+      // decide if double or tripple letter
+      if (this.word[i] === "{") {
+        scoreMultiplier = 2;
+      } else if (this.word[i] === "}") {
+        scoreMultiplier = 1;
+      } else if (this.word[i] === "[") {
+        scoreMultiplier = 3;
+      } else if (this.word[i] === "]") {
+        scoreMultiplier = 1;
```
PBara7a, line 31: You could consider improving this section to be able to score double/triple words with double/triple letters within. The way it is now, if you try to score {[w]ord}, once the ']' is detected your multiplier is set to 1 when the letters 'ord' should still be scored x2.

PBara7a, line 31: You could consider improving this section to be able to score double/triple words with double/triple letters within. The way it is now, if you try to score {[w]ord}, once the ']' is detected your multiplier is set to 1 when the letters 'ord' should still be scored x2.

PR #31, opened 2022-03-14
Tibor

Tibor22:


src/scrabble.js
```diff
@@ -1,6 +1,63 @@
...
+      value8: ['J', 'X'],
+      value10: ['Q', 'Z']
+    }
+  }
+
+  _validateString (word) {
+    if (typeof word !== 'string' || word.length === 0) {
+      return false
+    } else return true
+  }
```
qruz-hq, line 23: Is there a need to pass the `word` param into the function ? Since you're inside an object (`Scrabble`) you could have empty params and use `this.wordValue` instead of `word`.

Or maybe there's a reason for using params ?

src/scrabble.js
```diff
@@ -1,6 +1,63 @@
 class Scrabble {
   // Write your implementation here
 
+  constructor (wordValue) {
+    this.wordValue = wordValue
+    this.isDouble = false
+    this.isTriple = false
+    this.obj = {
```
Anicaaa, line 8: It's interesting. You used object to wrap the letters. I used the For Loop. Yours look easier and I'm going to try that next time I come across a similar exercise.

src/scrabble.js
```diff
@@ -1,6 +1,63 @@
...
+  _validateString (word) {
+    if (typeof word !== 'string' || word.length === 0) {
+      return false
+    } else return true
+  }
+
+  score () {
+    if (!this._validateString(this.wordValue)) return 0
+    let score = 0
+    this.wordValue = this.wordValue
```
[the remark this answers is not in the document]
Tibor22: yes, you're right, i should've remove it :D

src/scrabble.js
```diff
@@ -1,6 +1,63 @@
 class Scrabble {
```
Hoolio333, line 1: Not many pushes

src/scrabble.js
```diff
@@ -1,6 +1,63 @@
...
+  }
+
+  _validateString (word) {
+    if (typeof word !== 'string' || word.length === 0) {
+      return false
+    } else return true
+  }
+
+  score () {
+    if (!this._validateString(this.wordValue)) return 0
```
Hoolio333, line 26: !this - means not this?

src/scrabble.js
```diff
@@ -1,6 +1,63 @@
...
+    let score = 0
+    this.wordValue = this.wordValue
+      .toUpperCase()
+      .split('')
+      .forEach((letter) => {
+        for (const key of Object.keys(this.obj)) {
+          // Check for the brackets to decide if we need to double or triple up the scores
+          this._checkForDoubleOrTriple(letter)
```
Hoolio333, line 34: I like the clear descriptions - checksForDoubleOrTriple(letter)

src/scrabble.js
```diff
@@ -1,6 +1,63 @@
 class Scrabble {
   // Write your implementation here
 
+  constructor (wordValue) {
+    this.wordValue = wordValue
+    this.isDouble = false
```
Hoolio333, line 6: You don't need to put isDouble or isTriple in the constructor () ?

src/scrabble.js
Creatorrr1: Made easy and good for us 😁 but when the code is a little too much , more pushes is needed

src/scrabble.js
Creatorrr1: Nice this returns value zero for invalid strings like empty ones.

src/scrabble.js
```diff
@@ -1,6 +1,63 @@
...
+    if (letter === '{') {
+      this.isDouble = true
+    } else if (letter === '}') {
+      this.isDouble = false
+    } else if (letter === '[') {
+      this.isTriple = true
+    } else if (letter === ']') {
+      this.isTriple = false
+    }
+  }
```
Creatorrr1, line 56: NIce execution. Simple and clear code is the way 👌

src/scrabble.js
Creatorrr1: Nice use of arrays

Tibor22: meat to validate any string later come up as if the application could grow

PR #32, opened 2022-03-14
PULL request 1

Creatorrr1:


src/scrabble.js
```diff
@@ -1,6 +1,103 @@
...
+    if (this.word === null) {
+      return 0
+    }
+
+    
+    // Each individual letter (charAt) in the word is given point and added
+    // Return 0 if nothing is run
+    let totalpoints = 0
+    for (var i = 0; i < this.word.length; i++) {
+      totalpoints += letterScores[this.word.toLowerCase().charAt(i)] || 0;
```
qruz-hq: Nice and concise, we clearly understand how this system works, if the `letterScores` array returns `undefined` than we ad `0` to the score.

Hoolio333: I like this approach.

src/scrabble.js
```diff
@@ -48,56 +48,65 @@ class Scrabble {
...
+
+  /* PSEUDOCODE FOR NEXT OBJECTIVE STARTS */
+
+  // At default multiplier should be declared and assigned to 1
+  // Multiplier should be reassigned to 2 when a curly open bracket is detected
+  // The multiplier should be reassigned to 1 when a curly closed bracket is detected
+  // Multiplier should be assigned to 3 when open bracket is detected and reassigned to 1 when closing square bracket is detected
+  // The totalscores for the values inside will change due to the multipliers declared
+  // the totalscores at default (without the brackets) will not change
+
```
Hoolio333: I like the pseudo code. Breaking down the task & making it clear.

src/scrabble.js
```diff
@@ -57,55 +57,54 @@ class Scrabble {
...
+      } else if (letter === '}') {
+        multiplier = 1
+      } else if (letter === '[') {
+        multiplier = 3
+      } else if (letter === ']') {
+        multiplier = 1
+      } else {
+        const letterscore = letterScores[letter] || 0
+
+        totalpoints += letterscore * multiplier
```
Hoolio333, line 77: I like the approach to multiply the closing brackets by 1.

src/scrabble.js
Tibor22: this is a good way to go about solving the challenge

src/scrabble.js
```diff
@@ -1,6 +1,103 @@
+const Pointsystem = {
+  1: ['A' , 'a', 'E' , 'e', 'I', 'i', 'O','o', 'U','u', 'L','l', 'N','n', 'R','r', 'S','s', 'T','t'],
+  2: ['D','d', 'G','g'],
+  3: ['B','b', 'C','c', 'M','m', 'P','p'],
+  4: ['F','f', 'H','h', 'V','v', 'W','w', 'Y','y'],
+  5: ['K','k'],
+  8: ['J','j', 'X','x'],
+  10: ['Q','q', 'Z','z']
```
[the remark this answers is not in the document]
Anicaaa: Well done for using object and arrays. Very smart 🧠

src/scrabble.js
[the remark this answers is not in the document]
Creatorrr1: Yes ofcourse , this multplier technique is in a loop so when one bracket is detected then the multiplier will immediately especially when encapsulated.

PR #33, opened 2022-03-15
firstcommit

MikeCiccia:


src/scrabble.js
```diff
@@ -1,6 +1,73 @@
...
+
+    if (this.word === " \t\n") {
+      this.points += 0;
+      return 0;
+    }
+
+    if (this.word === null) {
+      this.points += 0;
+      return 0;
+    }
```
Alex90Jennings, line 22: could be replaced with if (this.word === " " || this.word === " \t\n" || this.word === null){return 0}

not a big deal, but can condense the code

src/scrabble.js
```diff
@@ -1,6 +1,73 @@
...
 
-module.exports = Scrabble
+    for (let i = 0; i < this.word.length; i++) {
+      const pointOne = ["a", "e", "i", "o", "u", "l", "n", "r", "s", "t"];
+      const pointTwo = ["d", "g"];
+      const pointThree = ["b", "c", "m", "p"];
+      const pointFour = ["f", "h", "v", "w", "y"];
+      const pointFive = ["k"];
+      const pointEight = ["j", "x"];
+      const pointTen = ["q", "z"];
```
Alex90Jennings, line 31: nicely named and well laid out

src/scrabble.js
```diff
@@ -1,6 +1,73 @@
...
+        pointTen.includes(this.word[i]) ||
+        pointTen.includes(this.word[i].toLowerCase())
+      ) {
+        this.points += 10;
+      }
+    }
+    return this.points;
+  }
+}
+module.exports = Scrabble;
```
Alex90Jennings, line 73: nicely structured and readable code

src/scrabble.js
```diff
@@ -1,6 +1,73 @@
...
+        pointEight.includes(this.word[i]) ||
+        pointEight.includes(this.word[i].toLowerCase())
+      ) {
+        this.points += 8;
+      } else if (
+        pointTen.includes(this.word[i]) ||
+        pointTen.includes(this.word[i].toLowerCase())
+      ) {
+        this.points += 10;
+      }
```
Alex90Jennings, line 68: you could include an else {this.points += 0}, this way if the user inputs an illegal character in the game çòàèù for example, it will score 0. With your current solution, it will return undefined and the whole score system will break.

This becomes important in the extended criteria, when the word includes [ { ] }

src/scrabble.js
0xtobycarlo, line 31: well laid out, easy to read

src/scrabble.js
sigristarisa: Also, what if this.word is other data types (e.g NaN, false etc)? It might pass the conditionals...
I would suggest doing `if(typeof this.word !== "string")`

src/scrabble.js
```diff
@@ -1,6 +1,73 @@
...
+    for (let i = 0; i < this.word.length; i++) {
+      const pointOne = ["a", "e", "i", "o", "u", "l", "n", "r", "s", "t"];
+      const pointTwo = ["d", "g"];
+      const pointThree = ["b", "c", "m", "p"];
+      const pointFour = ["f", "h", "v", "w", "y"];
+      const pointFive = ["k"];
+      const pointEight = ["j", "x"];
+      const pointTen = ["q", "z"];
+
+      if (
```
0xtobycarlo, line 33: for lines 33 to 68 an `else {this.points += 0}` is needed otherwise illegal/unidentified characters will not be scored and could break the system

src/scrabble.js
sigristarisa, line 68: You are repeating `toLowerCase()` to every letter, which could repetitive.
I would suggest making `the.word` into lowercase at the very beginning, and then looping through.
That way you could remove all the `this.word[i].toLowerCase()` in all the conditions:)

PR #41, opened 2022-07-22
Adds new solution to challenge

mbalabanov:


src/scrabble.js
```diff
@@ -1,5 +1,67 @@
...
+    Q: 10, Z: 10
+}
+
+let score = 0
+
+function scrabble(word) {
+    const wordArray = sanitizeAndConvert(word)
+    for (let letter of wordArray) { score += wordValues[letter] }
+    return outputFinalResultAndReset(score)
+}
```
vherus, line 18: Fantastic solution! But let's step away from the one liner ledge, you got a little carried away with it in this exercise

```js
function scrabble(word) {
    const wordArray = sanitizeAndConvert(word)
    
    for (let letter of wordArray) {
        score += wordValues[letter]
    }
    
    return outputFinalResultAndReset(score)
}
```

That takes up more space but it reads like a paragraph in a book, you can glance at it and the eyes know where to look

src/scrabble.js
```diff
@@ -1,5 +1,67 @@
...
+    return finalScore
+}
+
+function sanitizeAndConvert(word) {
+    if (word === null) { return ['_'] }
+    const wordCheckedForMultples = checkForMultiples(word);
+    const sanitizedWord = wordCheckedForMultples.toUpperCase().replace(' ', '_').replace('\t', '_').replace('\n', '_')
+    if (sanitizedWord.length === 1) { return [`${sanitizedWord}`] }
+    return sanitizedWord.split('')
+}
```
vherus, line 31:
```js
function sanitizeAndConvert(word) {
    if (word === null) {
        return ['_']
    }
    
    const wordCheckedForMultples = checkForMultiples(word);
    
    const sanitizedWord = wordCheckedForMultples.toUpperCase()
        .replace(' ', '_')
        .replace('\t', '_')
        .replace('\n', '_')
        
    if (sanitizedWord.length === 1) {
        return [`${sanitizedWord}`]
    }
    
    return sanitizedWord.split('')
}
```

Write code as if the person who's going to maintain it is homicidal and knows where you live :)

src/scrabble.js
```diff
@@ -1,5 +1,67 @@
...
+        return expandedWord.toUpperCase().replace('{', '').replace('}', '')
+    }
+    if(word.indexOf('[') !== -1) {
+        const position = word.indexOf('[') + 1
+        let letterToMultiply = word.slice(position, position + 1)
+        letterToMultiply = letterToMultiply + letterToMultiply + letterToMultiply
+        let expandedWord = [word.slice(0, position), letterToMultiply, word.slice(position)].join('');
+        return expandedWord.toUpperCase().replace('[', '').replace(']', '')
+    }
+    return word
```
vherus, line 64: This is a big one with very little white space! There's also some repetition going on, it could be a good chance to extract some things into their own functions.

Two if statements that check for the index of [ or {, the logic inside them do essentially the same thing but with different inputs. If you can identify the only pieces in there that are different, those are your inputs for some new functions

vherus (commented):
Great work!

src/scrabble.js
mbalabanov: Added white space (though not quite as much) and updated PR.

src/scrabble.js
mbalabanov: Added white space (though not quite as much) and updated PR.

src/scrabble.js
mbalabanov: Extracted functions from this large one. Updated PR

src/scrabble.js
mbalabanov: Added new solution based on today's morning session: new_scrabble.js (tests new_scrabble-spec.js).

src/scrabble.js
mbalabanov: Added new solution based on today's morning session: new_scrabble.js (tests new_scrabble-spec.js).

src/scrabble.js
mbalabanov: Added new solution based on today's morning session: new_scrabble.js (tests new_scrabble-spec.js).

PR #42, opened 2022-07-22
Scrabble with extensions. Also works for double/triple letters inside double/triple words.

AlexLind:


src/scrabble.js
```diff
@@ -1,5 +1,94 @@
...
+const pointValue = {
+  0: ['{', '}', '[', ']'],
+  1: ['a', 'e', 'i', 'o', 'u', 'l', 'n', 'r', 's', 't'],
+  2: ['d', 'g'],
+  3: ['b', 'c', 'm', 'p'],
+  4: ['f', 'h', 'v', 'w', 'y'],
+  5: ['k'],
+  8: ['j', 'x'],
+  10: ['q', 'z']
+}
```
vherus, line 39: Great approach

src/scrabble.js
```diff
@@ -1,5 +1,94 @@
...
+function assignPointsToWord(letter) {
+  return Number(
+    Object.keys(pointValue).find((k) => pointValue[k].includes(letter))
+  )
+}
+function isWordDouble(word, curlyCount) {
+  if (word[0] === '{' && word[word.length - 1] === '}' && curlyCount === 1) {
+    return true // IF condition for double word
+  }
+}
```
vherus, line 49: There's only one return here so this function will either return `true` or `undefined`, which is treated as falsy but it should probably return `false`

Or you could return the result of the evaluation :)

vherus (commented):
Great work!

PR #43, opened 2022-07-22
scrabble challenge finished

MattVwaves:


src/scrabble.js
```diff
@@ -1,5 +1,62 @@
-function scrabble() {
+aToT = ['a', 'e', 'i', 'o', 'u', 'l', 'n', 'r', 's', 't', 'A', 'E', 'I', 'O', 'U', 'L', 'N', 'R', 'S', 'T']
+dToG = ['d', 'g', 'D', 'G']
+bToP = ['b', 'c', 'm', 'p', 'B', 'C', 'M', 'P']
+fToY = ['f', 'h', 'v', 'w', 'y', 'F', 'H', 'V', 'W', 'Y']
+kay  = ['k', 'K']
+jToX = ['j', 'x', 'J', 'X']
+qToZ = ['q', 'z', 'Q', 'Z']
```
vherus, line 7: Remember to use the appropriate keyword when declaring variables, `const` in this case since the values won't change.

Additionally, an object might be better used here. Maybe something like

```js
const pointMap = {
  '1': ['a', 'e', 'i', 'o', 'u', 'l', 'n', 'r', 's', 't', 'A', 'E', 'I', 'O', 'U', 'L', 'N', 'R', 'S', 'T'],
  '2': ['d', 'g', 'D', 'G'],
  '3': ['b', 'c', 'm', 'p', 'B', 'C', 'M', 'P'],
  ...etc
}
```

In that example, the key on the object is the point value and the values of the keys are the arrays of characters associated with it. This might open you up to a cleaner solution in your logic

src/scrabble.js
```diff
@@ -1,5 +1,62 @@
...
+
+ 
+
+function scrabble(word){
+
+if (typeof word !== 'string'){
+        return 0
+    }    
+    
+total = 0
```
vherus, line 18: Keywords when declaring please :)

This one changes, so `let`

src/scrabble.js
```diff
@@ -1,5 +1,62 @@
...
+
+for (i = 0; i < word.length; i++){
+
+   
+    if (aToT.find(letter => letter === word[i])){
+        total = total + 1
+}
```
vherus, line 25: This could benefit from touching up the formatting, that closing curly bracket is trying to escape

PR #46, opened 2022-07-22
completed exercise

JamesB94:


src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+      addedTogether += 5
+    }
+    if (str.charAt(i) === 'J' || str.charAt(i) === 'X') {
+      addedTogether += 8
+    }
+    if (str.charAt(i) === 'Q' || str.charAt(i) === 'Z') {
+      addedTogether += 10
+    } 
+  }
+  return addedTogether
```
vherus, line 52: This is a valid approach and works fine! Those are some pretty bulky if statements though.

How about introducing an object that stores the point values for the letters? E.g.

```js
const pointMap = {
    A: 1,
    B: 3,
    etc...
}
```

Then you could use `Object.keys(pointMap)` to create an array of the object's keys, loop over the word to look at each character, and use square bracket syntax on the pointMap object to get the point value for that character

PR #50, opened 2022-09-08
George Roth

georgeroth:
Pull request so you can check my work for the Scrabble Challenge. Thanks!

### DIFF
--- a/src/scrabble.js
+++ b/src/scrabble.js
@@ -1,5 +1,85 @@
-function scrabble() {
+scrabble('{Pint}')
 
+function scrabble(word) {
+  // Catching the cheeky null value first
+  if (word === null) {
+    console.log('Why would you enter null?')
+    return 0
+  }
+  // If it's not null we move on
+  else {
+    const wordUpperCase = word.toUpperCase() // Coverting the given word to upper case
+    const wordSplit = wordUpperCase.split('') // Splitting the given word letter by letter
+    console.log(`Your letters are the following: ${wordSplit}`) // Displaying the letters to the user
+    let points = 0 // Setting up initial points tally at 0
+    for (const i of wordSplit) {
+      // Awarding points
+      if (
+        // letters for 1 point below
+        i === 'A' ||
+        i === 'E' ||
+        i === 'I' ||
+        i === 'O' ||
+        i === 'U' ||
+        i === 'L' ||
+        i === 'N' ||
+        i === 'R' ||
+        i === 'S' ||
+        i === 'T'
+      ) {
+        points += 1
+      } else if (
+        // letters for 2 points below
+        i === 'D' ||
+        i === 'G'
+      ) {
+        points += 2
+      } else if (
+        // letters for 3 points below
+        i === 'B' ||
+        i === 'C' ||
+        i === 'M' ||
+        i === 'P'
+      ) {
+        points += 3
+      } else if (
+        // letters for 4 points below
+        i === 'F' ||
+        i === 'H' ||
+        i === 'V' ||
+        i === 'W' ||
+        i === 'Y'
+      ) {
+        points += 4
+      } else if (
+        // letters for 5 points below
+        i === 'K'
+      ) {
+        points += 5
+      } else if (
+        // letters for 8 points below
+        i === 'J' ||
+        i === 'X'
+      ) {
+        points += 8
+      } else if (
+        // letters for 10 points below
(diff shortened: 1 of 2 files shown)


src/scrabble.js
```diff
@@ -1,5 +1,116 @@
...
+    console.log(`Why would you enter null? This gives you 0 points.`)
+    return 0
+  }
+
+  // If it's not null we move on
+  // Setting up initial points tally at 0
+  let points = 0
+
+  // Coverting the given word to upper case
+  const wordUpperCase = word.toUpperCase()
```
Joelzor, line 47: testing.... you could chain methods here like - word.toUpperCase().split(''). Doesn't make a whole lot of difference of course, just if you want less lines

src/scrabble.js
```diff
@@ -1,5 +1,116 @@
...
+  // Returning the final points tally
+  return points
+}
+
+function letterToPoints(letter) {
+  const value = pointsMap[letter] || 0
+  return value
+}
+
+function doubleAndTripleCharacterChecker(word, points) {
```
Joelzor, line 78: Very cool function, similar to mine but without the for loop. Awesome work 🥇

src/scrabble.js
georgeroth: Yeah, you're right, I may have gone overboard trying to make my thing super readable...

Joelzor: For readability you're probably correct in all fairness ;) I suppose chaining together a ton of methods would be pretty unreadable

georgeroth: Well, I think word.toUpperCase().split('') is probably readable even to us already and we have only been doing this for 3 weeks. Wonder what the best practice for this is, let's ask Ed

pseudocode.md
```diff
@@ -0,0 +1,58 @@
...
+- Inside the function we take one letter at a time
+- We create a new variable that takes the value from the object created in step one
+- We create a default value of 0 with the || condition when assigning the value to the new variable
+- Then we return the varible
+
+### Step 8
+
+TODO: we do some extra magic for the extensions
+
+### Step 9
```
Dex491, line 56: Love the use of MD elements in this file, makes the file look a lot nicer.

src/scrabble.js
```diff
@@ -1,5 +1,116 @@
...
+  // Checking for double and triple letters
+  points = doubleAndTripleCharacterChecker(wordSplit, points)
+
+  // Checking for double and triple words
+  points = doubleAndTripleWordChecker(wordSplit, points)
+
+  // Logging the final points tally to the console
+  console.log(`This gives you ${points} points in Scrabble!`)
+
+  // Returning the final points tally
```
Dex491, line 69: Separation of each process with their own comments makes the file really readable, if I were to nit-pick I'd say maybe too much? But that might just be a personal preference.

src/scrabble.js
```diff
@@ -1,5 +1,116 @@
...
+  const triplePowerLetterIndex = openingSquareBracketPlace + 1
+
+  if (closingCurlyBracketPlace - openingCurlyBracketPlace === 2) {
+    console.log(
+      `Double points for the letter "${word[doublePowerLetterIndex]}"!`
+    )
+    return (points += letterToPoints(word[doublePowerLetterIndex]))
+  } else if (closingSquareBracketPlace - openingSquareBracketPlace === 2) {
+    console.log(
+      `Triple points for the letter "${word[triplePowerLetterIndex]}"!`
```
Dex491, line 94: Really liking these logs, not only helps for testing but also provides a nice bit of user feedback.

src/scrabble.js
```diff
@@ -1,5 +1,116 @@
...
+    )
+    return (points += letterToPoints(word[triplePowerLetterIndex]) * 2)
+  } else {
+    return points
+  }
+}
+
+function doubleAndTripleWordChecker(word, points) {
```
Dex491, line 102: Having the separated functions really helps with readability, looks really concise. Personally I would have liked to see some comments here, but not major.

src/scrabble.js
georgeroth: Yeah I may have gone overboard...

src/scrabble.js
georgeroth: Great feedback, thanks Dave, I'll add some.

georgeroth: Added some!

PR #54, opened 2022-09-23
Scrabble challenge

wagnerAJS221:
Complete basic criteria of scrabble challenge completed and pseudo code updated. 
Extension crteria to be attempted and added.

### DIFF
--- a/src/scrabble.js
+++ b/src/scrabble.js
@@ -1,5 +1,44 @@
-function scrabble() {
+function scrabble(scrabbleWord) {
+  if (!scrabbleWord) return 0
+  const libary = {
+    a: 1,
+    e: 1,
+    i: 1,
+    o: 1,
+    u: 1,
+    l: 1,
+    n: 1,
+    r: 1,
+    s: 1,
+    t: 1,
+    d: 2,
+    g: 2,
+    b: 3,
+    c: 3,
+    m: 3,
+    p: 3,
+    f: 4,
+    h: 4,
+    v: 4,
+    w: 4,
+    y: 4,
+    k: 5,
+    j: 8,
+    x: 8,
+    q: 10,
+    z: 10
+  }
+  let sum = 0
+  scrabbleWord = scrabbleWord.toLowerCase()
 
+  for (let i = 0; i < scrabbleWord.length; i++) {
+    sum += libary[scrabbleWord[i]] || 0
+    // double or triple word to come here ( look at if statements )
+  }
+
+  return sum
 }
+
+console.log(scrabble(null))
 
 module.exports = scrabble
--- a/pseudocode
+++ b/pseudocode
@@ -1,0 +1,35 @@
+scrabble challenge pseudo code 
+
+understanding of challenge : 
+
+input - input a string value ( word)
+each letter in the english alphabet has a value preassigned to it already
+
+process - run the value through functions, loops and conditionals to them give me my output 
+charcters and null should return 0
+
+
+output - a numeric value based on the word.
+
+step 1 
+create my array with my letter scores in a let variable. 
+add guard feature so function will work with a non string null value. 
+
+
+step 2 
+create a loop that will fetch the letter values from the array. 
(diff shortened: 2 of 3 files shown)


src/scrabble.js
```diff
@@ -1,5 +1,58 @@
...
+    j: 8,
+    x: 8,
+    q: 10,
+    z: 10
+  }
+  let sum = 0
+  scrabbleWord = scrabbleWord.toLowerCase()
 
+  for (let i = 0; i < scrabbleWord.length; i++) {
+    sum += libary[scrabbleWord[i]] || 0
```
georgeroth, line 35: Nicely guarding against any random character with || 0

src/scrabble.js
```diff
@@ -1,5 +1,58 @@
-function scrabble() {
+function scrabble(scrabbleWord) {
+  if (!scrabbleWord) return 0
```
georgeroth, line 2: Nice and simple one line guard condition

src/scrabble.js
```diff
@@ -1,5 +1,58 @@
...
+  const libary = {
```
georgeroth, line 3: Might be a good idea to define library outside of the scope of scrabble()

src/scrabble.js
```diff
@@ -1,5 +1,58 @@
...
+  }
+  let sum = 0
+  scrabbleWord = scrabbleWord.toLowerCase()
 
+  for (let i = 0; i < scrabbleWord.length; i++) {
+    sum += libary[scrabbleWord[i]] || 0
+
+    // double or triple word to come here ( look at if statements )
+  }
+  if (
```
georgeroth, line 39: This could potentially be another function embedded into scrabble()?

src/scrabble.js
```diff
@@ -1,5 +1,58 @@
...
+    k: 5,
+    j: 8,
+    x: 8,
+    q: 10,
+    z: 10
+  }
+  let sum = 0
+  scrabbleWord = scrabbleWord.toLowerCase()
 
+  for (let i = 0; i < scrabbleWord.length; i++) {
```
georgeroth, line 34: A forEach loop may have been simpler here but a for loop is just as fine

src/scrabble.js
Joelzor: Seconded, you could write a separate function, passing in the score and then returning the new score with added bonus points. But very cool implementation

src/scrabble.js
SamuelOdigie, line 3: Declaring the const library at the top would give more visibility to the function scrabble

src/scrabble.js
```diff
@@ -1,5 +1,58 @@
...
+    scrabbleWord[scrabbleWord.length - 1] === ']'
+  ) {
+    return sum * 3
+  } else if (
+    scrabbleWord[0] === '{' &&
+    scrabbleWord[scrabbleWord.length - 1] === '}'
+  ) {
+    return sum * 2
+  }
+  return sum
```
SamuelOdigie, line 50: good use of if statements and using the loop function

src/scrabble.js
Haroogle, line 34: Nice use of a for loop

src/scrabble.js
Haroogle, line 39: remember to add a test for this in the spec file

src/scrabble.js
wagnerAJS221: could you please explain why ?

georgeroth: My thinking is that this way only your scrabble() function has access to the library. If you were to declare separate functions for the extensions besides scrabble(), like, let's say, tripleLetter(), and then call it inside scrabble(), tripleLetter() wouldn't be able to use your library because your library lives only inside scrabble(). For now, the way your code works it's completely fine though, that's why I was saying "might be" :)

wagnerAJS221: thank you for the explanation. That is a very good point. essentially my current code will then be a bit more limited than if it was done the otherway.
thank you

PR #56, opened 2022-09-25
exercise and extensions complete

Haroogle:


src/scrabble.js
```diff
@@ -1,5 +1,85 @@
-function scrabble() {
+const letterValue = {
```
dearshrewdwit, line 1: This is really interesting - how does this work with your main function?

src/scrabble.js
```diff
@@ -1,5 +1,85 @@
...
+  j: 8,
+  x: 8,
+  q: 10,
+  z: 10
+}
+const word = 'ho{t}DOg'
+function scrabble(word) {
+  let score = 0
+  if (word === null) {
+    // Guard condition for empty or null words
```
dearshrewdwit, line 33: I really like your use of comments to explain your code!

src/scrabble.js
```diff
@@ -1,5 +1,85 @@
...
+  h: 4,
+  v: 4,
+  w: 4,
+  y: 4,
+  k: 5,
+  j: 8,
+  x: 8,
+  q: 10,
+  z: 10
+}
```
dearshrewdwit, line 28: Have you thought about doing xyz?

src/scrabble.js
```diff
@@ -1,5 +1,85 @@
...
+      score += letterValue[characters[i + 1]] * 3
+      return score
+    }
+  }
+  return score
+}
 
+function scoreMultiplier(characters, score) {
+  if (
+    // Checks that the the whole word is contained within special brackets and not just the first and last characters
```
fatimaahmed14, line 64: did you write both these functions to only run with the word hot dog or will it run with any word when you change the const word?

src/scrabble.js
```diff
@@ -1,5 +1,85 @@
...
+  let score = 0
+  if (word === null) {
+    // Guard condition for empty or null words
+    return 0
+  }
+  const characters = word.toLowerCase().split('') // splits word into individual characters contained within an array, all lowercase
+  characters.forEach((character) => {
+    score += letterValue[character] || 0 // Checks each character within the array with its value in letterValue and adds that toscore. If its not there nothing is added
+  })
+  score = characterMultiplier(characters, score) // created separate function to check if charcter is within special brackets for a multiplier
```
fatimaahmed14, line 40: Implementing the 2 extension functions into the main function as opposed to adding more conditions makes your scrabble function look a lot more tidy and concise.

spec/scrabble-spec.js
```diff
@@ -34,4 +34,12 @@ describe("Scrabble", () => {
   it('scores a case-insensitive word', () => {
     expect(scrabble('OXYPHENBUTAZONE')).toEqual(41)
   })
+
+  it('scores include variables for triple and double letter bonuses', () => {
```
OlhaDorofiychuk, line 38: great job on adding tests

src/scrabble.js
OlhaDorofiychuk, line 64: clever expanded guard for characters

src/scrabble.js
```diff
@@ -1,5 +1,85 @@
...
+  w: 4,
+  y: 4,
+  k: 5,
+  j: 8,
+  x: 8,
+  q: 10,
+  z: 10
+}
+const word = 'ho{t}DOg'
+function scrabble(word) {
```
lucianosimoni, line 30: Good implementation of the variable. Seems to be more meaningful

src/scrabble.js
```diff
@@ -1,5 +1,85 @@
...
+  characters.forEach((character) => {
+    score += letterValue[character] || 0 // Checks each character within the array with its value in letterValue and adds that toscore. If its not there nothing is added
+  })
+  score = characterMultiplier(characters, score) // created separate function to check if charcter is within special brackets for a multiplier
+  score = scoreMultiplier(characters, score) // Created separate fucntion at the bottom to check for double and triple word multipliers
+  return score
+}
+console.log(scrabble(word))
+
+function characterMultiplier(characters, score) {
```
lucianosimoni, line 46: Nice one!
Another way of implementing this in fewer lines would be to move those codes inside the for loop you did to check every character, then those checks could store if the char is a special char into a global variable, and once you are checking every char: if the global var is true then this char would be special.
Just a quick logic 😅

src/scrabble.js
Haroogle: I tested it with different words yep

PR #57, opened 2022-09-26
Scrabble Challenge

mustafamuhith:
spent hours trying to figure it out but got there finally I think. I need more work on functions

### DIFF
--- a/src/scrabble.js
+++ b/src/scrabble.js
@@ -1,5 +1,42 @@
-function scrabble() {
-
+const alphabetData = {
+  a: 1,
+  e: 1,
+  i: 1,
+  o: 1,
+  u: 1,
+  l: 1,
+  n: 1,
+  r: 1,
+  s: 1,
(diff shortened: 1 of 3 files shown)


akibria12 (commented):
1. very clear and concise.
2. guard approach was interesting and using forEach method
3. writing a comment before each step/function would be great.

src/scrabble.js
```diff
@@ -1,5 +1,42 @@
...
+  z: 10
 }
 
+function scrabble(word) {
+  let score = 0
+  if (word === null) {
+    return 0
+  }
+  const alphaNum = word.toLowerCase().split('')
+  alphaNum.forEach((alphaNum) => {
```
Joelzor, line 36: You could maybe change the variable name inside forEach for easier readability- letter for example. Otherwise you have alphaNum twice referring to both the whole array + individual array items. But otherwise really neat and tidy implementation

PR #58, opened 2022-09-26
scrabble challenge 

fatimaahmed14:
- Core completed 
- extension 1 attempted but not as running code. It is all commented out

### DIFF
--- a/pseudocode.md
+++ b/pseudocode.md
@@ -3,9 +3,15 @@
 ## [input = a word]  =>  [output = score of word]
 
 step 1. Create a function that accepts the string as the input
+
 step 2. Create an object that associates the points to the corect (before the function)
+
 step 3. Within the function make letters in the string all the same case (toUppercase)
+
 step 4. Create a variable that will be used to store the outcome of the numbers added
+
 step 5. Use a loop that will go through the srting then use += to add number value to the variable created
+
 step 6. Add conditons regarding empty strings + nulls, these should be placed at the start of the function to ensure they run first
+
 step 7. Return the variable or 0
--- a/pseudocode.md
+++ b/pseudocode.md
@@ -2,16 +2,16 @@
 
 ## [input = a word]  =>  [output = score of word]
 
-step 1. Create a function that accepts the string as the input
+1. Create a function that accepts the string as the input
 
-step 2. Create an object that associates the points to the corect (before the function)
+2. Create an object that associates the points to the corect (before the function)
 
-step 3. Within the function make letters in the string all the same case (toUppercase)
+3. Within the function make letters in the string all the same case (toUppercase)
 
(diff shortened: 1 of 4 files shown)


src/tripleanddouble.js
```diff
@@ -0,0 +1,35 @@
...
+//     //     } else if (word[0] === '[' && word[word.length - 1] === ']') {
+//     //         return scrabble(word)* 3
+//     // } else {
+//     //     return scrabble(word)
+//     // 
+// }
+// }
+
+
+// function doubleWordScore(word) {
```
OlhaDorofiychuk, line 21: hello. Not sure why you don't consider this as a running code, in my opinion double and triple score for a whole word is correct..

pseudocode.md
```diff
@@ -0,0 +1,17 @@
+# Scrabble Challange
+
+## [input = a word]  =>  [output = score of word]
+
```
OlhaDorofiychuk, line 4: easy to understand pseudocode

src/scrabble.js
```diff
@@ -1,5 +1,47 @@
...
+    'V': 4,
+    'W': 4,
+    'X': 8,
+    'Y': 4,
+    'Z': 10 
+}
 
+function scrabble(word) {
+    let total = 0
+if (word === null){
```
lucianosimoni, line 32: This condition seems to be useless 🤔
Null is never returned in the test I did, even calling the function without an argument (the code just breaks).
A condition to avoid code break here would be a good replacement.

src/scrabble.js
```diff
@@ -1,5 +1,47 @@
...
+    return 0
+} else if (word === ''){
+    return 0
+} else if (word){
```
lucianosimoni, line 36: It could be just a normal call or maybe an else statement (not else if).
I tried with it just as a normal call and it worked fine

fatimaahmed14: So only use 1 if statement then leave the rest of the code in the functions curly brackets

src/tripleanddouble.js
fatimaahmed14: I tried to run both of the codes, the function with both double and triple and the 2 when they are separated, but they would constantly come up with an error. So I will try again today

src/scrabble.js
lucianosimoni: Yepss, but as long as the code works, it is all good😊

src/scrabble.js
lucianosimoni: > This condition seems to be useless 🤔 Null is never returned in the test I did, even calling the function without an argument (the code just breaks). A condition to avoid code break here would be a good replacement.

Naaah, wrong me 😅😅

fatimaahmed14:
Changed my core code slightly as one of the tests were failing  and also finished the extension

PR #59, opened 2022-09-26
Martin Lynch

mlync87:


src/scrabble.js
```diff
@@ -1,5 +1,54 @@
-function scrabble() {
+function scrabble(word) {
```
mustafamuhith, line 1: A declared function with the parameter of 'word'.

src/scrabble.js
```diff
@@ -1,5 +1,54 @@
...
+    y: 4,
+    k: 5,
+    j: 8,
+    x: 8,
+    q: 10,
+    z: 10
+  }
+  // #2 Write a guard function that will return the value '0' when input is null/not a previously stated character.
 
+  if (word == null || undefined) {
```
mustafamuhith, line 33: This makes sense and easy to understand

src/scrabble.js
```diff
@@ -1,5 +1,54 @@
...
+
+  let sum = 0
+  let i = 0
+
+  // #3 Create a variable that will contain the declared word and;
+  //  -put all characters in lower case
+  //  -divide it into an array.
+  word = word.toLowerCase()
+
+  // #4 Use a for loop to go through the array, and assign each item its assigned numerical value.
```
Dex491, line 45: Really liking the step-by-step comments in this, makes it really easy to read both function and intent of the code.

src/scrabble.js
```diff
@@ -1,5 +1,54 @@
-function scrabble() {
+function scrabble(word) {
+  //  Declare an object that shows all letters and their assigned numerical values.
+  const alphabet = {
```
mustafamuhith, line 3: Alphabet is declared within the function, this is interesting because I thought it was supposed to be outside.

src/scrabble.js
```diff
@@ -1,5 +1,54 @@
...
+  //  -divide it into an array.
+  word = word.toLowerCase()
+
+  // #4 Use a for loop to go through the array, and assign each item its assigned numerical value.
+  for (i = 0; i < word.length; i++) {
+    // #5 Calculate the sum of all items in array or return 0 if it is empty.
+    sum += alphabet[word[i]] || 0
+  }
+  // #5 Return the sum of items in the array.
+  return sum
```
mustafamuhith, line 51: Altogether looks like great work, especially the comments which showed your thought process. What may be an improvement for next time could be using a different method of presenting the code and that this may improve function-writing in general.

src/scrabble.js
```diff
@@ -1,5 +1,54 @@
...
+  let sum = 0
+  let i = 0
+
+  // #3 Create a variable that will contain the declared word and;
+  //  -put all characters in lower case
+  //  -divide it into an array.
+  word = word.toLowerCase()
+
+  // #4 Use a for loop to go through the array, and assign each item its assigned numerical value.
+  for (i = 0; i < word.length; i++) {
```
Dex491, line 46: Would have liked to see forEach here, but obviously this still works fine.

src/scrabble.js
mlync87: good spot! I'll change that this afternoon :)

src/scrabble.js
akibria12: forLoop() approach is very interesting here as to get the total score.

src/scrabble.js
akibria12: creating variable inside the function very interseting

src/scrabble.js
akibria12: very easy to understand overall as comment has been added in each step

PR #61, opened 2022-09-26
work

SamuelOdigie:
done

### DIFF
--- a/src/scrabble.js
+++ b/src/scrabble.js
@@ -1,5 +1,44 @@
-function scrabble() {
-
+const letternum = {
+  a: 1,
+  e: 1,
+  i: 1,
+  o: 1,
+  u: 1,
+  l: 1,
+  n: 1,
+  r: 1,
+  s: 1,
+  t: 1,
+  d: 2,
+  g: 2,
+  b: 3,
+  c: 3,
+  m: 3,
+  p: 3,
+  f: 4,
+  h: 4,
+  v: 4,
+  w: 4,
+  y: 4,
+  k: 5,
+  j: 8,
+  x: 8,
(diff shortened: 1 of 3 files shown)


src/scrabble.js
```diff
@@ -1,5 +1,44 @@
...
+  z: 10
+}
+
+function scrabble(word) {
+  if (word === null || word === '') return 0
+
+  let sum = 0
+
+  const upperCaseScore = word.toLowerCase().split('')
+  upperCaseScore.forEach((letter) => {
```
georgeroth, line 36: Nice use of forEach here

src/scrabble.js
```diff
@@ -1,5 +1,44 @@
...
+    sum += letternum[letter] || 0
```
georgeroth, line 37: Nicely guarding against any other random characters and spaces with || 0

src/scrabble.js
```diff
@@ -1,5 +1,44 @@
-function scrabble() {
+const letternum = {
```
georgeroth, line 1: It might be a good idea to declare the letternum variable outside of the scrabble() scope

src/scrabble.js
```diff
@@ -1,5 +1,44 @@
...
+  q: 10,
+  z: 10
+}
+
+function scrabble(word) {
+  if (word === null || word === '') return 0
+
+  let sum = 0
+
+  const upperCaseScore = word.toLowerCase().split('')
```
Joelzor, line 35: Would change upperCaseScore to lowerCaseScore. Or maybe even lowerCaseArray to be really explicit. Otherwise awesome :)

PR #62, opened 2022-09-26
Scrabble Challenge

Dex491:
Core and extension 1 completed, pseudocode for ext. 2 completed but code not fully implemented as I was pretty sick this weekend.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+    "prettier.useTabs": true,
+    "editor.tabSize": 2,
+    "prettier.tabWidth": 2,
+    "editor.detectIndentation": false
+}
--- a/src/letter-values.js
+++ b/src/letter-values.js
@@ -1,0 +1,30 @@
+const values = {
+  a: 1,
+  e: 1,
+  i: 1,
+  o: 1,
+  u: 1,
+  l: 1,
+  n: 1,
+  r: 1,
+  s: 1,
+  t: 1,
+  d: 2,
+  g: 2,
+  b: 3,
+  c: 3,
+  m: 3,
+  p: 3,
+  f: 4,
+  h: 4,
+  v: 4,
+  w: 4,
+  y: 4,
+  k: 5,
+  j: 8,
+  x: 8,
+  q: 10,
+  z: 10
+}
+
+module.exports = values
--- a/src/scrabble.js
+++ b/src/scrabble.js
@@ -4,21 +4,27 @@ function scrabble(word) {
   let points = 0
   if (word != null) {
     // safeguarding against no value
+    const character = word.split('')
+    // split the given word into individual characters in an array
     if (word.includes('{', '}')) {
       // if word contains brackets
       const firstBracket = word.indexOf('{') // creates variable that's the index of the first bracket location
       const secondBracket = word.indexOf('}') // same as above but for the second bracket
-      console.log(firstBracket)
       if (firstBracket !== word[0] && secondBracket !== word[word.length - 1]) {
         // checking if the brackets include the whole word, if so ignore this and continue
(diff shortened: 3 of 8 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "prettier.useTabs": true,
```
Joelzor, line 2: Already configuring the annoying Prettier, love to see it

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
-function scrabble() {
+const values = require('./letter-values.js') // fetch the values table
 
+function scrabble(word) {
+  let points = 0
+  if (word != null) {
```
mustafamuhith, line 5: Like the use of the exclamation in the guard, very easy to notice and read

src/letter-values.js
```diff
@@ -0,0 +1,30 @@
+const values = {
```
Joelzor, line 1: Nice use of modules to not clutter the main file

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
-function scrabble() {
+const values = require('./letter-values.js') // fetch the values table
```
Joelzor, line 2: I'd maybe add more empty space between some lines for greater readability but love the comments, helps to easily follow what's happening

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
 
+function scrabble(word) {
+  let points = 0
+  if (word != null) {
+    // safeguarding against no value
+    const character = word.split('')
+    // split the given word into individual characters in an array
+    if (word.includes('{', '}')) {
+      // if word contains brackets
+      const firstBracket = word.indexOf('{') // creates variable that's the index of the first bracket location
```
mustafamuhith, line 11: This variable is interesting because it is very specific and used only for one purpose.  I'm not sure, but could this have been made into one variable for both brackets? (overall thumbs up)

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+      if (
+        firstSqBracket !== word[0] &&
+        secondSqBracket !== word[word.length - 1]
+      ) {
+        // checking if the brackets include the whole word, if so ignore this and continue
+        const toBeTripled = character.splice(firstSqBracket, secondSqBracket) // create a variable that is the letters within the bracket
+        toBeTripled.forEach((tripledLetter) => {
+          points += values[tripledLetter.toLowerCase()] * 3 || 0
+          // change every letter to lowercase, match the letter with the letter and value in the values array and add to the points variable after multiplying by 3, or if it's not in the array, return 0 points
+        })
```
mustafamuhith, line 35: I now see how your working are with IF statements, it is separated and comprised methodically for all minute details.

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+      points *= 2
+      console.log('Extension 2')
+    }
+    if (word[0].includes('[') && word[word.length - 1].includes(']')) {
+      points *= 3
+      console.log('Extension 2')
+    }
+  }
+  return points
 }
```
mustafamuhith, line 53: Altogether, very interesting read. It mostly made sense to me and your work shows everything we have learnt in the past three three weeks implemented. What may be improved if you were to do this again, could be using a different method for the core.js so that you may increase in your dexterity and have a wider range and skill set.

akibria12 (commented):
1. adding plain -description.md file to explain the solution
2. export .js file to break down the step
3. extension work was interesting and wonder if it could be done more simplified approach.

src/scrabble.js
Dex491: It likely could have been, this was just my first implementation and I really should have gone back over it to see if it could be refined, thanks for the feedback.

PR #63, opened 2022-09-26
AbulKibria

akibria-dev:


src/scrabble.js
```diff
@@ -1,5 +1,46 @@
...
+function scrabble(word) {
+  let points = 0
+
+  if (word === null) {
+    return 0
+  }
+  const wordList = word.split('')
+  wordList.forEach((letter) => {
+    points += letterPoints[letter.toLowerCase()] || 0
+  })
```
OlhaDorofiychuk, line 39: I like your use of arrow function to minimalize characters

fatimaahmed14, line 39: a wonderful and concise function, I like the use of the forEach

PR #64, opened 2022-09-26
scrabble challenge

Yemaneab100:


src/psuedocode.md
```diff
@@ -2,5 +2,5 @@ Here are the steps needed to accomplish the task.
 
 1.  I will declare an abject that stores each letter by assigning a the given value called letterValues.
 2. declare a variable that stores the sum of each letter in a word called total.
-3.  using map function split the variable, and using conditinals statment, I will compute each letter a value from the letterValues object.
```
wagnerAJS221: would help to have your steps in normal numerical steps i.e. 1,2,3,4,5,6,7 etc etc

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+  J: 8,
+  X: 8,
+  Q: 10,
+  Z: 10
 }
 
+function scrabble(word) {
+  let total = 0
+  if (word === ' \t\n' || word === null) {
+    total = 0
```
wagnerAJS221, line 34: i think the code is precise and simple. easy to read

Yemaneab100: Thank you so much Alex.

src/scrabble.js
```diff
@@ -1,5 +1,46 @@
-function scrabble() {
+const { empty } = require("joi/lib/types/alternatives")
+let letters = {
```
SamuelOdigie, line 2: can use const as letters won't change

src/scrabble.js
```diff
@@ -1,5 +1,46 @@
...
+  Q: 10,
+  Z: 10
+}
+
+function scrabble(word) {
+  let total = 0
+  if (word === ' \t\n' || word === null) {
+    total = 0
+  } else {
+    for (let i = 0; i < word.length; i++) {
```
SamuelOdigie, line 36: nice use of loop function

src/scrabble.js
georgeroth: +1 to Sam's feedback here. Also, letters could be defined outside of scrabble() to make it a global variable, so if you start breaking down scrabble into smaller functions with the extensions in the future, those other functions will also have access to it.

src/scrabble.js
georgeroth: +1 to Alex here, nice and simple -- well done!

src/scrabble.js
```diff
@@ -1,5 +1,46 @@
-function scrabble() {
+const { empty } = require("joi/lib/types/alternatives")
```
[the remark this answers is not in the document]
Yemaneab100: lolll sorry. added without my consent. amzing for identifying it lolll

PR #69, opened 2022-11-23
Ahmed Missouri - Core Completed

ayMissouri:


src/scrabble.js
```diff
@@ -1,5 +1,57 @@
-function scrabble() {
+const values = {
```
JakeBurdass, line 1: Nice use of Objects!

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+    "X": 8,
+    "Y": 4,
+    "Z": 10
+}
 
+function scrabble(string) {
+if (string === null) {
+    return 0
+} else {
+    var regEx = /^[A-Za-z]+$/
```
JakeBurdass, line 34: No idea what this does, but well done for understanding this

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+    if(string.match(regEx)) { 
+        const word = string.toUpperCase()
+        let sum = 0
+        for (let i = 0; i < string.length; i++) {
+            let current = word[i]
+            sum += values[current]
```
JakeBurdass, line 40: I like the pulling the data from the object method!

JakeBurdass, line 40: I would like to know more about how you can match the keys from the object with the characters on the string, then return the value of the key

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+        }
+    return sum
+} else { 
+    return 0
+}
+}
 }
 
+console.log(scrabble(''))
```
JakeBurdass, line 49: Great work! Nothing that I think you could improve upon

src/scrabble.js
Cloudy-96: same! I also tried to do this and ended up having to do things the long way round

src/scrabble.js
Cloudy-96, line 34: good use of this! it has really made your code nice and concise

src/scrabble.js
Jenocke: I approve

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+console.log(scrabble(" \t\n"))
```
Cloudy-96, line 50: Well done!

src/scrabble.js
Jenocke: I like it too. (I used the same method ^^ )

Jenocke: by using value[letter], letter will be replaced by the current letter during runtime. So when the computer will read "A" for example, it'll be value[A] and will be able to access that value.

src/scrabble.js
Jenocke, line 49: Indeed, really nice work. Good luck on the extension

PR #70, opened 2022-11-23
Chris Sach Scrabble

ChrisJS90:


src/scrabble.js
```diff
@@ -1,5 +1,59 @@
...
+  for (let i = 0; i < word.length; i++) {
+    score += letterValue(word[i])
+    // console.log(score)
+  }
+  return score
+}
+
+// console.log(scrabble('cabbage'))
+
 module.exports = scrabble
```
WestyCodes, line 59: I will combine my questions here:
1)  On line 47 - you use instanceof String. - What are you looking for with this command? I can see it's like your checker to make sure the information in the `word` is right, but what specifically does that command do?

2) What made you decide on arrays for your points data?

3)  What is your plan for implementing the letter and word multipliers into this code?

WestyCodes, line 59: 3 things to improve - only if I HAD to be nitpicky:
1) I personally feel naming your data arrays with something more descriptive would be beneficial - althought I did just realise the number corresponds to the points value not the amount of arrays...

2) You could move your 'work horse' functions/data structures to another file and import them over to the scrabble file which just contains your scrabble function.

3) It would change your code A LOT - but I wonder if perhaps using an object for the data structure would be 'better' than an array.

WestyCodes (commented):
I will combine my questions here:

1) On line 47 - you use instanceof String. - What are you looking for with this command? I can see it's like your checker to make sure the information in the word is right, but what specifically does that command do?

2) What made you decide on arrays for your points data?

3) What is your plan for implementing the letter and word multipliers into this code?

3 things to improve - only if I HAD to be nitpicky:

1) I personally feel naming your data arrays with something more descriptive would be beneficial - althought I did just realise the number corresponds to the points value not the amount of arrays...

2) You could move your 'work horse' functions/data structures to another file and import them over to the scrabble file which just contains your scrabble function.

3) It would change your code A LOT - but I wonder if perhaps using an object for the data structure would be 'better' than an array.

src/scrabble.js
```diff
@@ -1,5 +1,59 @@
-function scrabble() {
+const array1 = ['A', 'E', 'I', 'O', 'U', 'L', 'N', 'R', 'S', 'T']
```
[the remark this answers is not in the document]
ayMissouri: I agree with this, its something I should have done instead of making an object that takes up like 27 lines

src/scrabble.js
```diff
@@ -1,5 +1,59 @@
...
 }
 
+// function multiLetter(word) {
+//   if (word.includes('{')) {
+//   }
+// }
+
+function scrabble(word) {
+  let score = 0
+  if (typeof word !== 'string' || word instanceof String) {
```
ayMissouri, line 47: I would like to know what instanceof does, sounds interesting.

src/scrabble.js
ayMissouri, line 59: clean and easily readable code, good job!

src/scrabble.js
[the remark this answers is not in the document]
Benjipictor: Also agree this is so neat

src/scrabble.js
```diff
@@ -1,5 +1,59 @@
...
+const array4 = ['F', 'H', 'V', 'W', 'Y']
+const array5 = ['K']
+const array8 = ['J', 'X']
+const array10 = ['Q', 'Z']
 
+function letterValue(letter) {
+  let value = 0
+  const letterUpper = letter.toUpperCase()
+  // console.log(letterUpper)
+  if (array1.includes(letterUpper)) {
```
Benjipictor, line 13: using array.includes vs what id did of applying if statements to each character in the string im a big fan of

src/scrabble.js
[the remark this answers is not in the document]
Benjipictor: also really like this good way to disqualify what the program cant run against

Benjipictor (commented):
Q1: how much did your final code look like what you imagined when you started.
enjoyed this as a good example cant think of many questions.

PR #71, opened 2022-11-23
Jake Burdass Scrabble!

JakeBurdass:


src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+        } else {
+            value.push(0)
+        }
+    }
+
+    for (j = 0; j < value.length; j++) {
+        finalSum += value[j]
+    }
+
+    console.log(`your word was ${userString}, you scored ${finalSum} points!`)
```
Cloudy-96, line 50: I like how this logs so that we know what the score/number at the end represents! Good idea.

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+        } else if (str[i] === 'F' || str[i] === 'H' || str[i] === 'V' || str[i] === 'W' || str[i] === 'Y') {
+            value.push(4)
+        } else if (str[i] === 'K') {
+            value.push(5)
+        } else if (str[i] === 'J' || str[i] === 'X') {
+            value.push(8)
+        } else if (str[i] === 'Q' || str[i] === 'Z') {
+            value.push(10)
+        } else {
+            value.push(0)
```
Cj-cheng, line 42: like the idea of push+(value), why i didn't think of that, smart move

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+// 2.) Create function with loads of if statements to get numerical value. e.g if B (index 1 in above array) return 3. 
+// 3.) Store this data somewhere (object maybe?)
+// 4.) Create a function that takes a string as an argument
+// 5.) For loop within function that loops round the length of the string
+// 6.) Within the for loop translate each index of the string to it's numerical value (how??? - Maybe function from point 2 (index[i]) *Need to think of scope here*
+// 7.) Store these values in array
+// 8.) Loop around array to add
+// 9.) Return value 
+
+const lettersArray = ['A','B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z']
```
Cloudy-96, line 14: This part seems like it ended up not being needed? I did a similar thing though so...

src/scrabble.js
Jenocke: Nice idea indeed

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+    for (j = 0; j < value.length; j++) {
+        finalSum += value[j]
+    }
+
+    console.log(`your word was ${userString}, you scored ${finalSum} points!`)
+    return finalSum
+}
 }
 
+scrabble('chicken nuggets')
```
Jenocke, line 55: I like this test :)

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+// 8.) Loop around array to add
+// 9.) Return value 
+
+const lettersArray = ['A','B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z']
+
+function scrabble(userString) {
+    if (userString === null) {
+        return 0
+    } else {
+
```
Cj-cheng, line 20: i didnt notice this until my spec gave me so many failures notification hahaha

src/scrabble.js
Jenocke: The array is indeed useless. That being said, another data structure could be used instead

src/scrabble.js
```diff
@@ -1,5 +1,57 @@
...
+    }
+
+    for (j = 0; j < value.length; j++) {
+        finalSum += value[j]
+    }
+
+    console.log(`your word was ${userString}, you scored ${finalSum} points!`)
+    return finalSum
+}
 }
```
Cj-cheng, line 53: was so confused how to add all value together hahaha 

in one-word, gorgeous work, inspire me a lot

src/scrabble.js
Jenocke: The problem with this push+value is that you have longer computing than just adding the score to finalSum since the computer needs to go through another loop that is not needed

PR #72, opened 2022-11-23
Kevin Goffin - scrabble

Jenocke:


src/scrabble.js
```diff
@@ -1,5 +1,76 @@
-function scrabble() {
+const createLetterValue = () => {
+  const letterScoreTable = {
+    ...Object.fromEntries(
+      ['A', 'E', 'I', 'O', 'U', 'L', 'N', 'R', 'S', 'T'].map((key) => [key, 1])
```
dhanwadepooja2688, line 4: Great work !! you use the  .map() function

src/scrabble.js
```diff
@@ -1,5 +1,76 @@
...
+    ...Object.fromEntries(['B', 'C', 'M', 'P'].map((key) => [key, 3])),
+    ...Object.fromEntries(['F', 'H', 'V', 'W', 'Y'].map((key) => [key, 4])),
+    K: 5,
+    J: 8,
+    X: 8,
+    Q: 10,
+    Z: 10
+  }
+  return letterScoreTable
+}
```
Cj-cheng, line 17: gorgeous work, how you use (). map function and key, totally blow my mind,

src/scrabble.js
```diff
@@ -1,5 +1,76 @@
...
 
+function scrabble(word) {
+  let wordScore = 0
+  const letterScoreTable = createLetterValue()
+  if (word === null) {
+    return 0
+  }
+  wordScore = manageWord(word, letterScoreTable)
+  return wordScore
 }
```
Cj-cheng, line 75: will try to read this later, too much to digest, you are PRO

src/scrabble.js
```diff
@@ -1,5 +1,76 @@
-function scrabble() {
+const createLetterValue = () => {
+  const letterScoreTable = {
+    ...Object.fromEntries(
```
Cj-cheng, line 3: ...object i need go to google mean what

src/scrabble.js
```diff
@@ -1,5 +1,76 @@
-function scrabble() {
+const createLetterValue = () => {
```
MaybeFreak, line 1: I like the idea of dynamically creating the object but I am unsure how easy it would be to maintain in the case changes need to be made. Personally I prefer having an object in a separate file that holds all the keys and values in case an adjustments would need to be made.

src/scrabble.js
```diff
@@ -1,5 +1,76 @@
...
+  return letterScoreTable
+}
+
+const getLetterScore = (letter, multiplier, letterScoreTable) => {
+  if (letterScoreTable[letter.toUpperCase()] !== undefined)
+    return letterScoreTable[letter.toUpperCase()] * multiplier
+  else return 0
+}
+
+const manageWord = (word, letterScoreTable) => {
```
dhanwadepooja2688, line 25: quick question on how u can pass  the object name (letterScoreTable ) in function parameter ?

src/scrabble.js
```diff
@@ -1,5 +1,76 @@
...
+  let curlyOpen = false
+  for (let i = 0; i < word.length; i++) {
+    if (
+      (word[i] === `}` && curlyOpen === false) ||
+      (word[i] === `]` && bracketOpen === false)
+    )
+      return 0
+
+    let letterValue = 0
+    switch (word[i]) {
```
MaybeFreak, line 38: Very interesting idea to use switch rather than if statements, definitely makes the code simpler and easier to read.

src/scrabble.js
MaybeFreak, line 75: Overall great code, super readability and very different (in a good way) than the way I approached this problem.

src/scrabble.js
Jenocke: This is what I do in a way. I'm just creating the object in a function

src/scrabble.js
Jenocke: the object is a type of variable in a way, so we can pass them as simple variables. Javascript will know that it's an object

PR #73, opened 2022-11-23
Scrabble challenge completed

Manojkumar883:
finished only src challenge , need to work on extension

### DIFF
--- a/src/scrabble.js
+++ b/src/scrabble.js
@@ -1,5 +1,50 @@
-function scrabble() {
-
+function scrabble(word) {
+  const lettersAN = {
+    A: 1,
+    B: 3,
+    C: 3,
+    D: 2,
+    E: 1,
+    F: 4,
+    G: 2,
+    H: 4,
+    I: 1,
+    J: 8,
+    K: 5,
+    L: 1,
+    M: 3,
+    N: 1,
+    O: 1,
+    P: 3,
+    Q: 10,
+    R: 1,
+    S: 1,
+    T: 1,
+    U: 1,
+    V: 4,
+    W: 4,
+    X: 8,
+    Y: 4,
+    Z: 10
+  }
+  let sum = 0
+  let letter = ''
+  if (word === '') {
+    return 0
+  }
+  if (word === ' \t\n') {
+    return 0
+  }
+  if (word === null) {
+    return 0
+  }
+  word = word.toUpperCase()
+  for (let i = 0; i < word.length; i++) {
+    letter = word[i]
+    sum += lettersAN[letter]
+  }
+  return sum
 }
-
+console.log(scrabble('CABBAGE'))
+console.log(scrabble('Manoj'))
 module.exports = scrabble
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ You'll need these:
 | ----                          |  ----  |
 | A, E, I, O, U, L, N, R, S, T  |     1  |
 | D, G                          |     2  |
-| , C, M, P                    |     3  |
+| B, C, M, P                    |     3  |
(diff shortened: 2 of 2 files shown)


src/scrabble.js
```diff
@@ -1,5 +1,50 @@
-function scrabble() {
-
+function scrabble(word) {
+  const lettersAN = {
```
Akis-Hajittofi, line 2: I see you've put 'AN' to indicate capital letters, but does 'AN' actually mean anything? Perhaps a better name could be something like 'lettersCapital'.

src/scrabble.js
```diff
@@ -1,5 +1,50 @@
...
+  if (word === ' \t\n') {
+    return 0
+  }
+  if (word === null) {
+    return 0
+  }
+  word = word.toUpperCase()
+  for (let i = 0; i < word.length; i++) {
+    letter = word[i]
+    sum += lettersAN[letter]
```
Akis-Hajittofi, line 49: Really nice way of calculating the score by giving each letter its score value in your object and then doing the calculation here.

src/scrabble.js
ChrisJS90, line 2: Nice use of an object, I didn't think to use an object like this. Quite clever!

src/scrabble.js
```diff
@@ -1,5 +1,50 @@
...
+    U: 1,
+    V: 4,
+    W: 4,
+    X: 8,
+    Y: 4,
+    Z: 10
+  }
+  let sum = 0
+  let letter = ''
+  if (word === '') {
```
Akis-Hajittofi, line 33: Good use of conditions to see if the word meets the requirements first before getting to the foor loop.

src/scrabble.js
```diff
@@ -1,5 +1,50 @@
...
+  }
+  if (word === ' \t\n') {
+    return 0
+  }
+  if (word === null) {
+    return 0
+  }
+  word = word.toUpperCase()
+  for (let i = 0; i < word.length; i++) {
+    letter = word[i]
```
Akis-Hajittofi, line 48: Since this is changing everytime the loop is run, it could also work by making it 'let letter = word[i]'?

src/scrabble.js
```diff
@@ -1,5 +1,50 @@
...
+    X: 8,
+    Y: 4,
+    Z: 10
+  }
+  let sum = 0
+  let letter = ''
+  if (word === '') {
+    return 0
+  }
+  if (word === ' \t\n') {
```
ChrisJS90, line 37: Interesting take on passing the tests!

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
-function scrabble() {
-
+function scrabble(word) {
+  const lettersAN = {
```
WestyCodes, line 2: Object use! Yeaaaah! I also like how you had it in alphabetical order... (rather than points order) much easier as a human to go through and find the relevant score for each letter.

src/scrabble.js
WestyCodes: This also caught my eye, as I'm not entirely sure what it does!

src/scrabble.js
WestyCodes: Also agree with Akis - its a very nice clean way to add the total together.

WestyCodes (commented):
Questions:

1) What does line 37 do? the ' \t\n' you have my curiosity!
2) No further questions!

Improvements (very nit picky):
1) You could have your data structure outside your function, so its always there and available.
2) Once you start implementing more functions, you could move your data structure and functions to a new file and use import/export module to have an even cleaner file.
3) Everything else seems great!

PR #74, opened 2022-11-23
Kat's Scrabble

Cloudy-96:


src/scrabble.js
```diff
@@ -1,5 +1,183 @@
-function scrabble() {
 
+//Core Scrabble: nailed it
+let newWordArr = []
+let sum = 0
+
+function scrabble(str) {
+    if (str === null) {
```
Jenocke, line 7: That's a nice idea to check if there's a word first. That way you can have faster computing and probably less errors

src/scrabble.js
```diff
@@ -1,5 +1,183 @@
-function scrabble() {
 
+//Core Scrabble: nailed it
+let newWordArr = []
+let sum = 0
```
Jenocke: Nice idea to have this variable outside the other functions. Be carefull about global variables tho, since it can be source of errors in more complex programs

src/scrabble.js
```diff
@@ -1,5 +1,183 @@
...
+        }
+
+
+    }
+    //console.log(newWordArr)
+
+    for (j = 0; j < newWordArr.length; j++) {
+        sum += newWordArr[j]
+        // console.log('inside internal loop', sum)
+    }
```
Jenocke, line 41: That loop is useless is you update score directly in the other loop

src/scrabble.js
```diff
@@ -1,5 +1,183 @@
...
+// Tried to make an object and use this in the code, but so far, no good
+const doublePointLetter = {
+2: '{A}' || '{E}' || '{I}' || '{O}' || '{U}'|| '{L}'|| '{N}' || '{R}' || '{S}' || '{T}' ,
+4: '{D}' || '{G}',
+6: '{B}' || '{C}' || '{M}' || '{P}',
+8: '{F}' || '{H}' || '{V}' || '{W}' || '{Y}',
+10: '{K}',
+16: '{J}' || '{X}',
+20: '{Q}' || '{Z}'
 }
```
Jenocke: An object is a good idea, I think. I would suggest to continue looking into it

src/scrabble.js
```diff
@@ -1,5 +1,183 @@
...
+            return 0
+        }
+
+
+    }
+    //console.log(newWordArr)
+
+    for (j = 0; j < newWordArr.length; j++) {
+        sum += newWordArr[j]
+        // console.log('inside internal loop', sum)
```
Cj-cheng, line 40: was so confused dont know how to add all the value together, hahaha , yours is fantastic,lets share some more after class

src/scrabble.js
```diff
@@ -1,5 +1,183 @@
...
+        } else if (newWord[i] === 'F' || newWord[i] === 'H' || newWord[i] === 'V' || newWord[i] === 'W' || newWord[i] === 'Y') {
+            newWordArr.push(4)
+        } else if (newWord[i] === 'K') {
+            newWordArr.push(5)
+        } else if (newWord[i] === 'J' || newWord[i] === 'X') {
+            newWordArr.push(8)
+        } else if (newWord[i] === 'Q' || newWord[i] === 'Z') {
+            newWordArr.push(10)
+        } 
+
```
Cj-cheng, line 29: was so lost in the OR function then struggling roughly like few hours, like push(value) idea, wise

src/scrabble.js
```diff
@@ -1,5 +1,183 @@
...
+// }
+
+// console.log(scrabble('d{o}g')) //should = more than 5
+
+// // Pseudo code for Whole word
+// // 1. make a function for double word: if '${[word]}' or '${{word}}'... x sum*2 (or *3)... gotta figure out the syntax
+
+// // clean code:
+// tried to start building functions separately for the extensions but it hasn't worked so far... defs a later problem
+module.exports = scrabble
```
Cj-cheng, line 183: YOU TRIED, OMG, you are amazing, like your encouragement, inspiring me A LOT

PR #75, opened 2022-11-23
Akis Hajittofi - Let's play Scrabble!

Akis-Hajittofi:


src/scrabble.js
```diff
@@ -1,5 +1,127 @@
-function scrabble() {
+function scrabble(string) {
+  // let score = 0
 
+  if (!string) {
```
ChrisJS90, line 4: Didn't know this was a way to check if it's a string! Great job!

src/scrabble.js
```diff
@@ -1,5 +1,127 @@
...
+  const scoreOfFive = 'K'
+  const scoreOfEight = ['J', 'X']
+  const scoreOfTen = ['Q', 'Z']
+
+  function calculateScore(word) {
+    let score = 0
+    for (let i = 0; i < word.length; i++) {
+      const counter = i
+
+      for (let j = 0; j < scoreOfOne.length; j++) {
```
ChrisJS90, line 23: Interesting take to run a loop within a loop to check if the letter is in the array. There is a built in function though that makes it a lot easier which is array.includes()

src/scrabble.js
```diff
@@ -1,5 +1,127 @@
...
+function scrabble(string) {
+  // let score = 0
 
+  if (!string) {
+    return 0
+  }
+
+  const word = string.toUpperCase()
+
+  const scoreOfOne = ['A', 'E', 'I', 'O', 'U', 'L', 'N', 'R', 'S', 'T']
```
ChrisJS90, line 10: Arrays! I took the same approach

src/scrabble.js
WestyCodes: Also didn't know - very clean, big fan!

src/scrabble.js
```diff
@@ -1,5 +1,127 @@
...
+  function doubleScore() {
+    for (let i = 0; i < word.length; i++) {
+      const counter = i
+      let charsToBeDoubled = '.'
+      if (word[counter] === '}') {
+        return null
+      }
+      if (word[counter] === '{') {
+        // while (word[counter] !== '}') {
+        for (let i = word.indexOf(word[counter]); i < word.length; i++) {
```
WestyCodes, line 71: Really interesting use of indexOf - cool!

src/scrabble.js
```diff
@@ -1,5 +1,127 @@
...
+        // console.log('INSIDE DOUBLE SCORE ---> ', charsToBeDoubled)
+        return null
+        // }
+      }
+    }
+    // score = 0
+    return 0
+  }
+
+  let finalScore = calculateScore(word) + doubleScore() + tripleScore()
```
WestyCodes, line 116: I like the use of multiple functions to come together to create the final score!

ChrisJS90 (commented):
Does your double and triple score work? Can't figure out how to copy it and try it on my machine haha

WestyCodes (commented):
Questions:
1) Are you using null to show an invalid string?
2) If yes to above - What made you decide to use null?
3) With each extension implementation, did you run into any problems?

Improvements:
1) You could move your functions to another file, and use the import/export module so your scrabble function is just calling a couple functions to get to an answer

2) It may be easier to use a .includes() method or something similar, rather than having to have nested loops to go through your data arrays.

3) I personally liked using an object for the Point Values data structure - may be an improvement... but it may not be...

src/scrabble.js
```diff
@@ -1,5 +1,127 @@
...
+      for (let j = 0; j < scoreOfTen.length; j++) {
+        if (word[counter] === scoreOfTen[j]) {
+          score += 10
+        }
+      }
+      // console.log(`${counter} and score is: ${score}`)
+    }
+    return score
+  }
+
```
Benjipictor, line 61: that's impressive the way you got the nested for loop to work for you! love that man.

src/scrabble.js
```diff
@@ -1,5 +1,127 @@
...
+  const word = string.toUpperCase()
+
+  const scoreOfOne = ['A', 'E', 'I', 'O', 'U', 'L', 'N', 'R', 'S', 'T']
+  const scoreOfTwo = ['D', 'G']
+  const scoreOfThree = ['B', 'C', 'M', 'P']
+  const scoreOfFour = ['F', 'H', 'V', 'W', 'Y']
+  const scoreOfFive = 'K'
+  const scoreOfEight = ['J', 'X']
+  const scoreOfTen = ['Q', 'Z']
+
```
Benjipictor, line 17: good use of arrays keeps everything looking much cleaner

PR #76, opened 2022-11-23
first attempt CJ

Cj-cheng:


src/scrabble.js
```diff
@@ -1,5 +1,49 @@
-function scrabble() {
-
+function scrabble(string) {
+  if (string === null) {
+    return 0
+  } else {
+    const word = string.toUpperCase()
+    let value = 0
+    for (let i = 0; i <= word.length; i++) {
+      if (
```
MaybeFreak, line 8: You could create an object that holds all of the characters and their specific values, that way you only need to check if the character at index i is included in the object and if true you could then add that characters value to the value with something like `value += objectname[word[i]]`

src/scrabble.js
```diff
@@ -1,5 +1,49 @@
-function scrabble() {
-
+function scrabble(string) {
+  if (string === null) {
+    return 0
+  } else {
+    const word = string.toUpperCase()
```
MaybeFreak, line 5: Like that you transformed the string to Uppercase, find it easier to identify uppercase letters.

src/scrabble.js
```diff
@@ -1,5 +1,49 @@
...
+      if (
+        word[i] === 'A' ||
+        word[i] === 'E' ||
+        word[i] === 'I' ||
+        word[i] === 'O' ||
+        word[i] === 'U' ||
+        word[i] === 'L' ||
+        word[i] === 'N' ||
+        word[i] === 'R' ||
+        word[i] === 'S' ||
```
Hamza789987, line 17: could u have put them in same line to make it more easier to understand

src/scrabble.js
```diff
@@ -1,5 +1,49 @@
...
+        value += 2
+      } else if (
+        word[i] === 'B' ||
+        word[i] === 'C' ||
+        word[i] === 'M' ||
+        word[i] === 'P'
+      ) {
+        value += 3
+      } else if (
+        word[i] === 'F' ||
```
Hamza789987, line 31: much neater and nice when u added them up to see what is being returned and what code is executed after an if statement

src/scrabble.js
```diff
@@ -1,5 +1,49 @@
...
+        word[i] === 'V' ||
+        word[i] === 'Y'
+      ) {
+        value += 4
+      } else if (word[i] === 'K') {
+        value += 5
+      } else if (word[i] === 'J' || word[i] === 'X') {
+        value += 8
+      } else if (word[i] === 'Q' || word[i] === 'Z') {
+        value += 10
```
Hamza789987, line 42: u could make a space between each new score to separate the letters from different points

PR #77, opened 2022-11-23
core done

Benjipictor:


src/scrabble.js
```diff
@@ -1,5 +1,53 @@
-function scrabble() {
+let score = 0
 
+function findValues(word) {
+    let value1 = 0
+    // console.log(`what is i: ${i}`)
```
JakeBurdass, line 5: Great use of console.log to get clarity in each part of the program!

src/scrabble.js
```diff
@@ -1,5 +1,53 @@
-function scrabble() {
+let score = 0
 
+function findValues(word) {
```
JakeBurdass, line 3: Nice use of using two functions. We struggled with that in some pair work done previously so good to see you've learnt from that.

src/scrabble.js
```diff
@@ -1,5 +1,53 @@
...
+    let value1 = 0
+    // console.log(`what is i: ${i}`)
+    let comparisonLetter = word.charAt(i)
+    
+    if (comparisonLetter === 'A'||comparisonLetter === 'E'||comparisonLetter === 'I'||comparisonLetter === 'O'||comparisonLetter === 'U'||comparisonLetter === 'L'||
```
ayMissouri, line 8: I like that you managed to do it without an object

src/scrabble.js
```diff
@@ -1,5 +1,53 @@
-function scrabble() {
+let score = 0
 
+function findValues(word) {
+    let value1 = 0
+    // console.log(`what is i: ${i}`)
+    let comparisonLetter = word.charAt(i)
```
ayMissouri, line 6: I like that this index links to the for loop thats at the bottom instead of making a new loop for it

src/scrabble.js
```diff
@@ -1,5 +1,53 @@
...
+      findValues(word)
+    } 
+    let finalScore = score
+    score = 0
+    return finalScore
+    }
+
+// console.log(scrabble('a'))
+
 module.exports = scrabble
```
ayMissouri, line 53: One way i can think to improve this would be to see if you can use a function to scan through the letters instead of a bunch of if statements, other than that good job! 👍

src/scrabble.js
```diff
@@ -1,5 +1,53 @@
...
+        value1 = 8
+    }
+    else if (comparisonLetter == 'Q'||comparisonLetter == 'Z'){
+        value1 = 10
+    }
+    
+    return score += value1
 }
 
+function scrabble(str) {
```
Cloudy-96, line 35: I like that you made two separate functions so that each one has a clear role

Cloudy-96: it makes your code very readable! 10/10

ayMissouri (commented):
Good job.

src/scrabble.js
```diff
@@ -1,5 +1,53 @@
...
+    if (str === null||str === " \t\n" ){
+        return 0
+    }
+    if (str === ''){
+        return 0
+    }
+    let word = str.toUpperCase()
+    for (i = 0; i <= word.length; i++){
```
JakeBurdass, line 43: This is cool. Using i within another function then passing it to the function above. Good job

PR #78, opened 2022-11-24
Basic done pooja

dhanwadepooja2688:


src/scrabble.js
```diff
@@ -1,5 +1,51 @@
-function scrabble() {
-
+const lettervalue = {
```
MaybeFreak, line 1: I like that you use an object to hold all the values for the letters, this makes statements a lot more readable and the code generally more flexible.

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
...
+  world = world.toLowerCase()
+  if (world === ' \t\n') {
+    return 0
+  }
+  if (world === '') {
+    return 0
+  }
+
+  for (let i = 0; i < world.length; i++) {
+    letter = world[i]
```
MaybeFreak, line 45: Instead of assigning `world[i]` to a variable you could just use it directly in the line below, or maybe just declare the variable inside of the for loop to avoid possible problems since the variable is in a higher scope than it needs to be.

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
...
 }
 
+function scrabble(world) {
+  let sum = 0
+  let letter = ''
+  if (world === null) {
+    return 0
+  }
+  world = world.toLowerCase()
+  if (world === ' \t\n') {
```
MaybeFreak, line 37: creating another object to hold invalid characters, just like you did for the letters, could cut down on the amount of if statements you need to write and can make the code more flexible since you could avoid hardcoding statements.

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
...
+  h: 4,
+  v: 4,
+  w: 4,
+  y: 4,
+  k: 5,
+  j: 8,
+  x: 8,
+  q: 10,
+  z: 10
 }
```
Hamza789987, line 28: nice how the list is used to allow to see the points for each letter

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
...
+  }
+
+  for (let i = 0; i < world.length; i++) {
+    letter = world[i]
+    sum = sum + lettervalue[letter]
+  }
+  return sum
+}
+console.log(scrabble('quirky'))
 module.exports = scrabble
```
Hamza789987, line 51: nice how u finished the core with less lines of code

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
...
+    return 0
+  }
+  world = world.toLowerCase()
+  if (world === ' \t\n') {
+    return 0
+  }
+  if (world === '') {
+    return 0
+  }
+
```
Hamza789987, line 43: making spaces after doing short lines of code is not necessary and can make it a bit confusing

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
...
+  for (let i = 0; i < world.length; i++) {
+    letter = world[i]
+    sum = sum + lettervalue[letter]
```
Hamza789987, line 46: could u have made this more easier to use when updating, what are u updating

PR #79, opened 2022-11-24
done core

Hamza789987:


src/scrabble.js
```diff
@@ -1,5 +1,82 @@
-function scrabble() {
-
+function scrabble(str) {
+  let counter = 0 // must be local to restart and redefine back to 0 when finished otherwise global scope will add it up
+  if (!str || str.length === 0) {
+    return 0
+  }
+  for (let i = 0; i < str.length; i++) {
```
Manojkumar883, line 6: good use of if statement in the for loop and assigning each index value of string.

src/scrabble.js
```diff
@@ -1,5 +1,82 @@
-function scrabble() {
-
+function scrabble(str) {
+  let counter = 0 // must be local to restart and redefine back to 0 when finished otherwise global scope will add it up
+  if (!str || str.length === 0) {
```
Akis-Hajittofi, line 3: Very nice way of verifying if the string is empty! Also I think it could also work with just simply (!str)

src/scrabble.js
```diff
@@ -1,5 +1,82 @@
...
+    } else if (
+      str[i] === 'q' ||
+      str[i] === 'z' ||
+      str[i] === 'Q' ||
+      str[i] === 'Z'
+    ) {
+      counter += 10
+    } else return 0
+  }
+  return counter
```
Manojkumar883, line 79: great work and easy to understand, please check once did you add the values of null and '\t\n' should return 0

src/scrabble.js
ChrisJS90, line 3: Love this check for if it's a string (or empty!)

src/scrabble.js
```diff
@@ -1,5 +1,82 @@
-function scrabble() {
-
+function scrabble(str) {
+  let counter = 0 // must be local to restart and redefine back to 0 when finished otherwise global scope will add it up
+  if (!str || str.length === 0) {
+    return 0
+  }
+  for (let i = 0; i < str.length; i++) {
+    if (
+      str[i] === 'a' ||
```
ChrisJS90, line 8: Great job on getting all these comparisons in and looking nice! I took a different approach by creating arrays for each score value and each array including the letters associated with that score. But this works brilliantly!

src/scrabble.js
ChrisJS90, line 79: All in all this is very clean and easy to read. Great work!

src/scrabble.js
Akis-Hajittofi, line 6: Interesting way to check the letters. It may look long but it's simple and works fine!

src/scrabble.js
```diff
@@ -1,5 +1,82 @@
...
+    ) {
+      counter += 8
+    } else if (
+      str[i] === 'q' ||
+      str[i] === 'z' ||
+      str[i] === 'Q' ||
+      str[i] === 'Z'
+    ) {
+      counter += 10
+    } else return 0
```
Akis-Hajittofi, line 77: Good use of this else statement to return 0

Akis-Hajittofi (commented):
Good work Hamza!

PR #81, opened 2022-12-02
Scrabble challenge

Nozie6:


src/scrabble.js
```diff
@@ -1,5 +1,47 @@
...
+}
+
+function scrabble(anyword) {
+  if (anyword === null) {
+    return 0
+  }
+
+  let score = 0
+
+  const trimmedWord = anyword.trim().toUpperCase()
```
NoroAxper, line 37: .trim is a new function i had not seen before, nice find!

src/scrabble.js
```diff
@@ -1,5 +1,47 @@
...
+  for (let i = 0; i < trimmedWord.length; i++) {
+    const letter = trimmedWord[i]
+
+    score += scoreMap[letter]
+  }
 
+  return score
 }
```
NoroAxper, line 45: Over all very clean code, easy to understand and read. Awesome work!

src/scrabble.js
```diff
@@ -1,5 +1,47 @@
...
+  Y: 4,
+  K: 5,
+  J: 8,
+  X: 8,
+  Q: 10,
+  Z: 10
+}
+
+function scrabble(anyword) {
+  if (anyword === null) {
```
NoroAxper, line 31: Alternatively, I think you could put "anyword !== 'string'" and cover more anomalies such as if the input was a number?. (not 100% sure, maybe note even 80% just trying to think outside the box 🤪)

PR #84, opened 2023-03-15
Core Scrabble Completed

davidbrockway:
Core Scrabble Completed - Test Passed

### DIFF
--- a/src/scrabble.js
+++ b/src/scrabble.js
@@ -1,5 +1,77 @@
-function scrabble() {
+const scoreVaules = {
+    // Corresponding alphabet to letter score values
+    A: 1,
+    E: 1,
+    I: 1,
+    O: 1,
+    U: 1,
+    L: 1,
+    N: 1,
+    R: 1,
+    S: 1, 
+    T: 1,
+    D: 2,
+    G: 2,
+    B: 3,
+    C: 3,
+    M: 3,
+    P: 3,
+    F: 4,
+    H: 4,
+    V: 4,
+    W: 4,
+    Y: 4,
+    K: 5,
+    J: 8,
+    X: 8,
+    Q: 10,
+    Z: 10,
+
 
 }
+
+// console.log(scoreVaules) // Testing Purpose
+
+function scrabble(word) {
+    if (word === null) {
+        return 0;
+    }
+    word = word.trim().toLowerCase();
+    if (!word) {
+        return 0;
+    }
+    let s = 0; // score count 
+    for (let i = 0; i < word.length; i++ ){
+        letterr = word[i].toUpperCase();
+         s += parseInt(scoreVaules[letterr]);
+
+    }
+    return (s);
+
+}
+
+console.log(scrabble("OXYPHENBUTAZONE"));
+
+console.log(scrabble("OzYPHENaUTAZONE"));
+
+
+scrabble('') // should return 0
+
+scrabble(" \t\n") // should return 0
+
+scrabble(null) // should return 0
+
+scrabble('a') // should return 1
+
+scrabble('f') // should return 4
+
+scrabble('NaN') // should return 41
+
+scrabble('street') // should return 6
+
+scrabble('quirky') // should return 22
+
+scrabble('OXYPHENBUTAZONE') // should return 41
+
 
 module.exports = scrabble


src/scrabble.js
```diff
@@ -1,5 +1,98 @@
-function scrabble() {
+const scoreValues = {
```
marc-alzapiedi, line 1: I like the use of the object to assign the letters a score. Looking at other peoples code a lot of people took this approach. Its very neat, The way I did it is a bit more obnoxious.

src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+  Z: 10,
+}
 
+function isValidInput(word) {
+  return (word !== undefined && word !== null);
 }
 
-module.exports = scrabble
+function scrabble(word) {
+  if (!isValidInput(word)) {
```
marc-alzapiedi, line 36: I had to think the use of the bang on lines 32 and 36 but after thinking it through it makes sense. Clever!

src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+    return 0;
+  }
+
+  let s = 0; // stands for score
```
marc-alzapiedi, line 40: Personally I would avoid defining the variable like you have in this way where the variable is a single letter with a comment on what it means. 

It may be better to directly define the variable (score).

src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+function scrabble(word) {
+  if (!isValidInput(word)) {
+    return 0;
+  }
+
+  let s = 0; // stands for score 
+  let wordMultiplier = 1;
+  let i = 0;
+
+  while (i < word.length) {
```
marc-alzapiedi, line 44: love the while loop, very clever!

src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+    let letter = word[i].toUpperCase();
+
+    if (letter === '{' || letter === '[') {
+      let closingChar = letter === '{' ? '}' : ']';
+      let j = word.indexOf(closingChar, i + 1);
+      if (j === -1) {
```
marc-alzapiedi, line 50: Looking at this if statement seems counterintuitive first glance, but I think I know the trick your using. Also on line 48 it seems like you are checking to see if there are any characters that were included in the function argument that are not accepted, like (\ | % £ $).

src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+  K: 5,
+  J: 8,
+  X: 8,
+  Q: 10,
+  Z: 10,
+}
 
+function isValidInput(word) {
+  return (word !== undefined && word !== null);
 }
```
alexjshaw, line 33: nice clean way of exiting the scrabble function if a word isn't entered.

src/scrabble.js
alexjshaw, line 40: As Marc said, use more descriptive variable names! You only have to type them the first time, after that you can autofill them so there's no excuse for being lazy!

src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+
+  let s = 0; // stands for score 
+  let wordMultiplier = 1;
+  let i = 0;
+
+  while (i < word.length) {
+    let letter = word[i].toUpperCase();
+
+    if (letter === '{' || letter === '[') {
+      let closingChar = letter === '{' ? '}' : ']';
```
alexjshaw, line 48: Could you explain line 48, I don't quite understand what the === '{' ? '}' : ']' part is actually doing.

src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+      i++;
+    } else {
+      return 0;
+    }
+  }
+
+  return s * wordMultiplier;
+
+}
+
```
alexjshaw, line 94: This whole scrabble function seems a bit bloated. Breaking it down in to smaller functions would make it easier to see what each part is doing and check if individual sections are working correctly. Would help to isolate problems and tackle one issue at a time rather than trying to fix every issue at the same time.

src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+      i = j + 1;
+    } else if (letter === '|') {
+      let j = word.indexOf('|', i + 1);
+      if (j === -1 || (j - i) % 2 === 0) {
+        return 0;
+      }
+      let letters = word.substring(i + 1, j).toUpperCase();
+      let letterMultiplier = 1;
+      for (let k = 0; k < letters.length; k++) {
+        let letterScore = scoreValues[letters[k].toUpperCase()] || 0;
```
alexjshaw, line 69: If you're getting to the point of using k as a variable then you're probably going a bit too deep with the nested loops.

alexjshaw (commented):
Good effort overall, some nice logic at points but I think you've confused yourself a bit at times. Loops within loops within loops are always going to be hard to follow, and trying to break some of that out to its own functions might have made it easier to see what was working and what wasn't. Good effort though David.

src/scrabble.js
ManNavic, line 48: Have the same question as Alex

PR #86, opened 2023-03-15
Finished scrabble standart criteria

carolarruda:


src/scrabble.js
```diff
@@ -1,5 +1,191 @@
...
+        newWord[i].includes('I') ||
+        newWord[i].includes('O') ||
+        newWord[i].includes('U') ||
+        newWord[i].includes('L') ||
+        newWord[i].includes('N') ||
+        newWord[i].includes('R') ||
+        newWord[i].includes('S') ||
+        newWord[i].includes('T') ||
+        newWord[i].includes('a') ||
+        newWord[i].includes('e') ||
```
marc-alzapiedi, line 21: I think you can avoid rewriting the letters again in lowercase by writing on line 5:
const newWord = aWord.toUpperCase()

Then any string that is passed in the argument  regardless of whether it has 1 upper case or all lowercase will all be returned uppercase.

alexjshaw: What Marc said! Instead of having to declare values for the uppercase and lowercase letters you could just transform your entered word in to upper or lower case and assigned it to a new constant.

src/scrabble.js
```diff
@@ -1,5 +1,191 @@
...
+      ) {
+        sum += 10
+      }
+    }
+    return sum
+  }
+}
+
+function doubleLetter(aWord) {
+  const regex1 = /{[A-Z|a-z]}/g
```
alexjshaw, line 92: Really nice idea to use regex for your double and triple letter checks. Really clever and much more efficient than my method of using if loops. Great idea.

src/scrabble.js
```diff
@@ -1,5 +1,191 @@
...
+    if (
+      (aWord.includes('[') && aWord.includes(']')) ||
+      (aWord.includes('{') && aWord.includes('}'))
+    ) {
+      const finalScore =
+        normalLetter(aWord) + doubleLetter(aWord) + trippleLetter(aWord)
+
+      return finalScore
+    }
+    if (
```
marc-alzapiedi, line 172: Also clever way to check that there is never only one [ in the string and that there is always two, and there is no invalid characters. However, if there are characters entered like ? $ % / £ €, or other characters that shouldn't be included I didn't see a way for the program to return 0.

src/scrabble.js
```diff
@@ -1,5 +1,191 @@
-function scrabble() {
+function normalScore(aWord) {
+  if (aWord === '' || aWord === ' \t\n' || aWord === null) {
```
marc-alzapiedi, line 2: Does this check for invalid characters in the argument? like: % & $ £ € = / \

src/scrabble.js
```diff
@@ -1,5 +1,191 @@
...
+console.log('Double letter test', doubleLetter('{A}{A}'))
+
+function trippleLetter(aWord) {
+  const regex2 = /\[[A-Z | a-z]]/g
+  const found2 = aWord.match(regex2)
+  const trippleLetter = normalScore(found2) * 3
+  return trippleLetter
+}
+
+function normalLetter(aWord) {
```
alexjshaw, line 107: I'm not 100% sure on what this function does, from line 107 to 113. Any chance you can reply to this comment and explain it for me?

src/scrabble.js
```diff
@@ -1,5 +1,191 @@
...
+  const split = aWord.replaceAll(/{[A-Z | a-z]}/g, '')
+  const split2 = split.replaceAll(/\[[A-Z | a-z]]/g, '')
 
+  console.log('split3', split2.toString())
+  console.log('First Normal Score', normalScore(split2))
+  return normalScore(split2)
 }
 
+function doubleWords(aWord5) {
+  const myLength = aWord5.length
```
alexjshaw, line 117: Why does this suddenly change from aWord to aWord5?

src/scrabble.js
```diff
@@ -1,5 +1,191 @@
...
+  }
+}
+
+console.log('Double word score', doubleWords('{AAAAA}'))
+
+function scrabble(aWord) {
+  if (typeof aWord === 'string') {
+    const myLength = aWord.length
+    if (aWord[0] === '[' && aWord[myLength - 1] === ']') {
+      const newWord2 = aWord.slice(1, myLength - 1)
```
alexjshaw, line 132: Nice job using slice to extract the word from inside the brackets.

alexjshaw (commented):
I really like what you've done here.

Diving in to regex is intimidating but some really clever use of it, and I really like that you've got your double and triple letter/word functions broken out and just pushing their output in to a variable to be used in your main scrabble function. Makes it easier to read.

Great job :)

src/scrabble.js
carolarruda: Good point. I added most of them now but backslash will need a regex probably because it keeps reading it as escaping.

src/scrabble.js
carolarruda: This was to clean up the argument and to avoid double scoring.

line 108 I look for anything inside { } and replace with ' ' and the same for the letters in the squares brackets[ ]

So for example in the word H{A}NDLE it would look for the {A} in line 108 and replace it with empty.

then return the score for normal letters using a clean argument of HNDLE.

src/scrabble.js
carolarruda: Honestly, I was so desperate to get the test passed that I just tried changing that variable name. Thanks for spotting that. Will change it.

PR #87, opened 2023-03-15
rasheed adesokan => solution to scrabble challenge

desokan:


src/scrabble.js
JayVith, line 1: I like how neatly, you've laid everything out. It shows your thought processes very clearly

src/scrabble.js
```diff
@@ -1,5 +1,38 @@
...
+    // the inner for loop will iterate the each unique group
+    for (let j = 0; j < alphaGroup.length; j++) {
+      // the if condition is checking if each element of the alphaGroup array contains the current scrabble word element.
+      if (alphaGroup[j].includes(scrabbleWord[i])) {
+        // If the above condition is met, the scrabbleScore variable is updated
+        // with the letter value of the group.
+        scrabbleScore += alphaGroup[j][alphaGroup[j].length - 1]
+      }
+    }
+  }
```
procrastian, line 83: really good and clear comments here explaining your code - very nice to follow and understand :)

src/scrabble.js
```diff
@@ -1,38 +1,124 @@
...
+    for (let i = 0; i < doubleTheWord.length; i++) {
+      for (let j = 0; j < alphaGroup.length; j++) {
+        if (alphaGroup[j].includes(doubleTheWord[i])) {
+          scrabbleScore += alphaGroup[j][alphaGroup[j].length - 1] * 2
+        }
+      }
+    }
+  }
+
+  if (scrabbleWord.includes('[') && scrabbleWord.includes(']')) {
```
procrastian, line 60: I like the tidiness of checking that the [ ] have been properly interred in order to apply the multipliers!

src/scrabble.js
```diff
@@ -1,5 +1,124 @@
...
+  }
+
+  if (scrabbleWord.includes('{') && scrabbleWord.includes('}')) {
+    const startIndex = scrabbleWord.indexOf('{')
+    const endIndex = scrabbleWord.indexOf('}')
+    doubleTheWord = scrabbleWord.splice(
+      startIndex + 1,
+      endIndex - (startIndex + 1)
+    )
+
```
JayVith, line 50: I like the use of splice  but with the second position, would it also work if you just had simply endIndex - 1?

src/scrabble.js
```diff
@@ -1,38 +1,124 @@
...
+  if (scrabbleWord.includes('[')) {
+    if (scrabbleWord.indexOf(']') === -1) {
+      return scrabbleScore
+    }
+  }
+
+  if (scrabbleWord.includes(']')) {
+    if (scrabbleWord.indexOf('[') === -1) {
+      return scrabbleScore
+    }
```
procrastian, line 40: Perhaps a function could work here to check for each of { } [ ] rather than the same code being repeated?

procrastian (commented):
Lovely clean and easy to follow code Rasheed! Nice work! :)

src/scrabble.js
desokan: > Perhaps a function could work here to check for each of { } [ ] rather than the same code being repeated?

I agree. I will probably refactor the code and implement your suggestion. Thanks.

PR #88, opened 2023-03-15
Alistair Henshaw

Alistair1080:


src/scrabble.js
```diff
@@ -1,5 +1,146 @@
...
+  } else if (l === 'F' || l === 'H' || l === 'V' || l === 'W' || l === 'Y') {
+    return 4
+  } else if (l === 'K') {
+    return 5
+  } else if (l === 'J' || l === 'X') {
+    return 8
+  } else if (l === 'Q' || l === 'Z') {
+    return 10
+  }
+}
```
Rebsonau, line 89: Nice use of else if statements. I used a switch statement for this portion, you should check them out and see if it's something you'd like to utilise in the future.

src/scrabble.js
```diff
@@ -1,5 +1,146 @@
...
+      score += scoring(letter)
+    }
+    if (doubleArray.length !== 0) {
+      console.log(doubleArray)
+      for (let i = 0; i < doubleArray.length; i++) {
+        const letter = doubleArray[i].toUpperCase()
+        score += scoring(letter) * 2
+      }
+    }
+    if (tripleArray.length !== 0) {
```
carolarruda, line 50: Have you tried making this a function and testing it to work independently before including it in the primary function scrabble?

src/scrabble.js
```diff
@@ -1,5 +1,146 @@
-function scrabble() {
+const doubleArray = []
+const tripleArray = []
+const alphabet = [
```
carolarruda, line 3: Really liked the new const for the alphabet approach!

src/scrabble.js
```diff
@@ -1,5 +1,146 @@
...
+}
+
+function wordCheck(word) {
+  if (typeof word !== 'string') {
+    return false
+  } else {
+    for (let i = 0; i < word.length; i++) {
+      const character = word[i]
+      const c = character.toUpperCase()
+      if (alphabet.includes(c) === true) {
```
Rebsonau, line 98: Haven't seen this method before, Will have to check it out for myself

src/scrabble.js
```diff
@@ -1,5 +1,146 @@
...
+    }
+    if (tripleArray.length !== 0) {
+      console.log(tripleArray)
+      for (let i = 0; i < tripleArray.length; i++) {
+        const letter = tripleArray[i].toUpperCase()
+        score += scoring(letter) * 3
+      }
+    }
+  }
+  return score
```
Rebsonau, line 58: I can see what you're trying to do here, I went for the unorthodox method of pushing the value surrounded by brackets onto the end of an array I created for the word (twice if it was square brackets), thus duplicating it so that it recieved a double score or triple score respectively.

Rebsonau (commented):
Good job Alistair

PR #89, opened 2023-03-15
Core + Extension (attempted) - Korey

rockettdev:


src/scrabble.js
```diff
@@ -1,5 +1,137 @@
-function scrabble() {
+function isAlpha(str) {
+  return (
+    /^[A-Za-z]*$/.test(str) ||
```
desokan, line 3: Nice use of regex. Can you explain what the **/^[A-Za-z]*$/.test(str)** is meant to do? Thanks.

src/scrabble.js
```diff
@@ -1,5 +1,137 @@
...
+        }
+        count += tempCount
+      } else return 0
+    }
+  }
+  return count
+}
+
+console.log(scrabble('d{og'))
+
```
desokan, line 136: Generally, I like how you used call back and higher-order functions in your program. It makes your code reusable. It can also potentially make it adaptable to other projects. Well done.

Do you plan to make the three remaining edge cases pass the test? Good luck with that.

src/scrabble.js
```diff
@@ -1,5 +1,137 @@
...
+    str === '[' ||
+    str === ']'
+  )
+}
+
+function getLettersInBracket(fromIndex, toIndex, scrabbleArray) {
+  return scrabbleArray.slice(fromIndex, toIndex)
+}
 
+function getTotalScore(k, scrabbleArray) {
```
procrastian, line 15: Was there a reason for using the parameter of k for your getTotalScore function? It can get a bit confusing later on as to whether it is the letter 'k' or the parameter k that is being passed in

src/scrabble.js
```diff
@@ -1,5 +1,137 @@
...
+    return 5
+  }
+  if (scrabbleArray[k] === 'j' || scrabbleArray[k] === 'x') {
+    return 8
+  }
+  if (scrabbleArray[k] === 'q' || scrabbleArray[k] === 'z') {
+    return 10
+  }
+  return 0
 }
```
procrastian, line 61: Nice way to organise the reference letter scores, and give a return of 0 if an invalid character in given at the same time :)

src/scrabble.js
```diff
@@ -1,5 +1,137 @@
...
+          return score
+        }
+      }
+
+      if (isAlpha(scrabbleArray[k])) {
+        tempCount += getTotalScore(k, scrabbleArray)
+        if (
+          (scrabbleArray[k - 1] === '{' && scrabbleArray[k + 1] !== '}') ||
+          (scrabbleArray[k - 1] !== '{' && scrabbleArray[k + 1] === '}')
+        ) {
```
procrastian, line 112: nice way to ensure the brackets have been applied correctly!

procrastian (commented):
Nice job Korey! Was able to follow the logic through nicely, and good use of outside knowledge/research with the regex!
Using the parameter k to represent the letter you're looking for confused me a bit at the start, but I can see how it works.
Good luck finishing the extension!

src/scrabble.js
rockettdev: In essence, it checks from a range of A-Z characters and a-z also for case sensitive edge cases

src/scrabble.js
rockettdev: Yeah, that was probably something i should change as it made sense in my last iteration

src/scrabble.js
rockettdev: Thanks!

PR #90, opened 2023-03-15
Gideon Usenbor

gid-ctrl:


src/scrabble.js
```diff
@@ -1,5 +1,53 @@
...
+  if (word === null) return 0
+  for (let i = 0; i < word.length; i++) {
+    const letter = word[i].toUpperCase()
+    if (
+      scrabbleScores === '' ||
+      scrabbleScores === `\t` ||
+      scrabbleScores === `\n`
+    )
+      return 0
+    if (word === null) return 0
```
sarajbeazley, line 41: on lines 31 & 40, I see that you have repeated the same if statement with a return of 0 - is this necessary?

PR #92, opened 2023-03-16
Gabriel Rosengren

Gabriel-Rosengren:


src/scrabble.js
```diff
@@ -1,5 +1,109 @@
...
+  Z: 10
+}
+// Object containing each type of bracket as the key, whose value is another object containing information about said bracket 
+// (multiplier value, corresponding open/closing bracket, and what type of bracket it is (open or closing))
+const brackets = { 
+  '{': {multi: 2, corresponding: '}', type: 'open'},
+  '[': {multi: 3, corresponding: ']', type: 'open'},
+  '}': {multi: 2, corresponding: '{', type: 'close'},
+  ']': {multi: 3, corresponding: '[', type: 'close'}
+}
```
Rebsonau, line 36: This is new to me! Will definately have to look into this. Very clean

src/scrabble.js
```diff
@@ -1,5 +1,109 @@
...
+  H: 4,
+  V: 4,
+  W: 4,
+  Y: 4,
+  K: 5,
+  J: 8,
+  X: 8,
+  Q: 10,
+  Z: 10
+}
```
Rebsonau, line 28: Really like this approach

src/scrabble.js
```diff
@@ -1,5 +1,109 @@
...
+      multiplier = calculateMultiplier(str[i], stack, multiplier)
+      if (multiplier === -1) {
+        return 0
+      }
+
+      // if false - add the current letters corresponding value in the lettersToPoints object
+    } else {
+      let currentLetter = str[i].toUpperCase()
+
+      if (!(currentLetter in lettersToPoints)) {
```
Rebsonau, line 65: Like this, haven't seen it before

src/scrabble.js
```diff
@@ -1,5 +1,109 @@
...
+function calculateMultiplier(char, arr, multiplier) {
+  // switch statement to check if the current bracket we are looking at is a closing or opening bracket
+  switch(brackets[char].type) {
+    // if "char" is an opening bracket we return the corresponding value from the brackets object as the multiplier
+    case 'open':
+      arr.push(char)
+      return multiplier *= brackets[char].multi
+    // if "char" is an closing bracket we return multiplier with the corresponding value from the brackets object removed
+    case 'close':
+      // console.log("corresponding", brackets[char].corresponding)
```
Rebsonau, line 94: Can you confirm that this works as it is intended with not having a break after the first case

Rebsonau (commented):
Really nice work, great use of comments to explain everything.

src/scrabble.js
Gabriel-Rosengren: Yeah I used to have the breaks in there, but since we are using "return" to get out of the function entirely, it won't continue down the line!

src/scrabble.js
timmyctt: After reading this , I think i will start using more objects to code!

PR #97, opened 2023-03-17
Scrabble Challenge (WIP) Priya

ps975076:


src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+
+// Creating an if statement - that if a word has a length of 0 then this should return a score of 0
+
+// If words are entered in capitals then need to assign the toUpperCase() method to string
+
+function scrabble(word) {
+  if (word === null && word === '\n' && word === '\t' && word.length === 0) {
+    return 0
+  }
+}
```
sarajbeazley, line 23: i think this is the right idea for these special characters but a for loop needs to come after the if statement to iterate through each letter, adding together the values of each letter

sarajbeazley: so if....
 else... for loop

src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+  X: 8,
+  Q: 10,
+  Z: 10
+}
+let totalScore = 0
+for (let character of word) {
+  totalScore += letterScore[character] 
 }
+return totalScore;
+
```
sarajbeazley, line 66: i think this should come first at the top of the code - sort of sets the scene and then other functions come below. its important to return the value of each letter so that the info is held and letter values can be added together.
Also, look up charAt() method

src/scrabble.js
```diff
@@ -1,5 +1,99 @@
...
+  H: 4,
+  V: 4,
+  W: 4,
+  Y: 4,
+  K: 5,
+  J: 8,
+  X: 8,
+  Q: 10,
+  Z: 10
+}
```
ample-samples, line 61: The letterScore object takes up a lot of space inside of the function, can you put it outside of the function?

src/scrabble.js
```diff
@@ -1,5 +1,99 @@
...
+// For e.g,
+// console.log(greeting);
+// Expected output: "   Hello world!   ";
+
+// console.log(greeting.trim());
+// Expected output: "Hello world!";
+
+// Creating an if statement - that if a word has a length of 0 then this should return a score of 0
+
+// If words are entered in capitals then need to assign the toUpperCase() method to string because it becomes case sensitive.
```
ample-samples, line 17: I like your thought process!

ample-samples (commented):
A little hard to read but otherwise good!

PR #99, opened 2023-03-17
Mantas Navickis [work in progress]

ManNavic:


src/scrabble.js
```diff
@@ -1,5 +1,67 @@
...
+    Z: 10
+  }
+  if ((!word, typeof word !== 'string')) {
+    return 0
+  }
+  let points = 0
+  for (let x = 0; x < word.length; x++) {
+    const letters = word[x].toUpperCase()
+    // console.log(letters)
+    if (word === '' || word === '\t' || word === '\n' || word === null) {
```
NoroAxper, line 45: Should the 'word' be changed to 'letters' since you are assigning each index of 'word' to = letters?

src/scrabble.js
```diff
@@ -1,5 +1,67 @@
-function scrabble() {
+// 1.Create a score chart
+// 2.Create a if statment that returns 0 if
+// scrabble('') // should return 0
+// scrabble(" \t\n") // should return 0
+// scrabble(null) // should return 0
+// 3. Somehow loop throught the word and count the letters
+// 4.Count the points
 
+function scrabble(word) {
```
Shylan21, line 9: Are you going straight for the word? If you're trying to compare each letter to a value I think that 'word' maybe should change to 'letters' instead?

PR #100, opened 2023-03-17
Marc Alzapiedi

marc-alzapiedi:


src/scrabble.js
```diff
@@ -1,5 +1,172 @@
...
+    ) {
+      charPoints *= 2
+    }
+
+    if (
+      caseSensitivity.search(/\[/) === 0 &&
+      caseSensitivity.search(/\]/) === caseSensitivity.length - 1
+    ) {
+      charPoints *= 3
+    }
```
alexjshaw, line 134: Nice use of .search and a bit of regex.

src/scrabble.js
```diff
@@ -1,5 +1,172 @@
...
+      '{',
+      '}',
+      '[',
+      ']',
+      ' ',
+      '',
+      '.'
+    ]
+    return acceptableChars.filter((pass) => pass.includes(charExtract))
+  }
```
alexjshaw, line 49: I like the use of filter to extract just the valid characters from the input string.

src/scrabble.js
```diff
@@ -1,5 +1,172 @@
...
+      charExtract === 'O' ||
+      charExtract === 'U' ||
+      charExtract === 'L' ||
+      charExtract === 'N' ||
+      charExtract === 'R' ||
+      charExtract === 'S' ||
+      charExtract === 'T'
+    ) {
+      charPoints = 1
+    }
```
ManNavic, line 65: Different way to add value to the letters  👍

src/scrabble.js
```diff
@@ -1,5 +1,172 @@
...
+      triplePoints += 1
+    }
+
+    if (!incorrectTokenControl().includes(charExtract)) {
+      return 0
+    }
+
+    scrabbleScore += charPoints
+  }
+
```
alexjshaw, line 150: It feels like this huge If statement could be broken down in to multiple smaller statements or functions to make it more readable. Rather than having one for loop that checks for letter scores and whether it's a double or triple letter could be split up in to different areas of the code.

src/scrabble.js
```diff
@@ -1,5 +1,172 @@
...
+    caseSensitivity.search(/\]/) === caseSensitivity.length - 2 &&
+    caseSensitivity.search('}') === caseSensitivity.length - 1
+  ) {
+    return scrabbleScore * 3
+  }
+
+  return scrabbleScore
 }
 
 module.exports = scrabble
```
alexjshaw, line 172: One thing I'm not quite picking up when reading your code is how exactly it's checking for the following:

1) Cases where it's a double and triple word score at the same time.
2) How it's checking for edge cases such as {d}o{g}

Overall I think that breaking your code up a bit more in to different functions would make it easier to follow. Good job though.

alexjshaw (commented):
Good job getting it all done. I've left a question about how a couple of bits of your code work at the bottom. If you'd be happy to give me an answer to those at some point I'd be interested to understand your solution better.

src/scrabble.js
marc-alzapiedi: Screenshot 2023-03-17 144948
Hi Alex, so for the cases where it's a double or triple word, the picture shows the code that executes that. So if you look in the condition of the if statement, I use .search() to check if one of [ or { is at the beginning of each string. I thought it would always be in the indexed position of 0 or 1 regardless of the argument input in the function. Then to seek for ] or } I figured it would always be at the end of whatever caseSensitivity.length() - 1, and this ensures that regardless of the argument, the indexed position would always be found using .length(). 

image
So for the edge cases, this is the code that checks for that. So since I put this in the loop, each time the loop iterates and it extracts a character from the argument and it passes it in the if statement condition if (caseSensitivity.includes(`{${charExtract}}`)). This checks the entire string if the current character that was extracted in the loop is inside  {} or [] regardless of how many times they appear. i.e ({d}{o}{g} or {d}o{g} or d{o}g)

If you have any more questions write to me!
I hope I explained myself well enough and also I hope I understood and answered your questions clearly. Let me know!

marc-alzapiedi: Screenshot 2023-03-17 144948
Hi Alex, so for the cases where it's a double or triple word, the picture shows the code that executes that. So if you look in the condition of the if statement, I use .search() to check if one of [ or { is at the beginning of each string. I thought it would always be in the indexed position of 0 or 1 regardless of the argument input in the function. Then to seek for ] or } I figured it would always be at the end of whatever caseSensitivity.length() - 1, and this ensures that regardless of the argument, the indexed position would always be found using .length().

Screenshot 2023-03-17 145721
So for the edge cases, this is the code that checks for that. So since I put this in the loop, each time the loop iterates and it extracts a character from the argument and it passes it in the if statement condition if (caseSensitivity.includes({${charExtract}})). This checks the entire string if the current character that was extracted in the loop is inside {} or [] regardless of how many times they appear. i.e ({d}{o}{g} or {d}o{g} or d{o}g)

If you have any more questions write to me!
I hope I explained myself well enough and also I hope I understood and answered your questions clearly. Let me know!

PR #101, opened 2023-03-17
Jay Vithani

JayVith:
Core completed - working on extension still


src/scrabble.js
```diff
@@ -1,5 +1,73 @@
...
+  u: 1,
+  v: 4,
+  w: 4,
+  x: 8,
+  y: 4,
+  z: 10
+}
+
+function scrabble(anyWord) {
+  if (
```
Gabriel-Rosengren, line 32: I like the indents on the if statement, makes it very readable and easy to understand

src/scrabble.js
```diff
@@ -1,5 +1,73 @@
...
+  const word = anyWord.toLowerCase()
+
+  let preScore = 0
+
+  for (let i = 0; i < word.length; i++) {
+    const basicScore = letterValues[word[i]]
 
+    if (basicScore > 0) {
+      preScore += basicScore
+    }
```
NoroAxper, line 48: You could make this a bit cleaner by taking out the if statement and directly start adding since all your letterValues are already more than 0

src/scrabble.js
```diff
@@ -1,5 +1,73 @@
...
+  }
+  return preScore
```
NoroAxper, line 50: Overall very easy to read, understand and follow, good job bro!!

PR #150, opened 2023-09-13
Kye

yee0802:


src/scrabble.js
```diff
@@ -2,31 +2,80 @@ const { letters } = require('./letter.js')
```
Chloe070196, line 2: First things first - your commit title is incredibly relatable ^^

src/scrabble.js
```diff
@@ -2,31 +2,80 @@ const { letters } = require('./letter.js')
...
   }
 
   const stringContainsNum = checkString(word)
 
-  if (typeof word === 'string' && stringContainsNum === false) {
+  // Multiple checks for valid input
+  if (
+    typeof word === 'string' &&
+    stringContainsNum === false &&
+    /^[a-zA-Z{}[\]]*$/.test(word)
```
Chloe070196, line 21: Looks good! I think I've just seen Nazar use sth similar in his code (.test() method) - how does it work?

src/scrabble.js
```diff
@@ -2,31 +2,80 @@ const { letters } = require('./letter.js')
...
+  ) {
     for (let i = 0; i < word.length; i++) {
-      scrabbleArr.push(word[i].toLowerCase())
+      const letter = word[i].toLowerCase()
+
+      // Checking for ({,[), if true then multiplier applied
```
Chloe070196, line 26: I like how you've solved the extension - I'd never have thought to include a clause to reset the multiplier to 1, that seems like a really smart way to go about this!

src/scrabble.js
yee0802: It's new knowledge to me! 
Here's the MDN definition: 
The test() method of RegEx instances executes a search with this regular expression for a match between a regular expression and a specified string. Returns true if there is a match; false otherwise.

src/scrabble.js
```diff
@@ -1,5 +1,81 @@
...
+const scrabble = (word) => {
+  let output = 0
+  let mulitplier = 1
+  // Arrays to test for valid input
+  const checkerArr = []
+  const checkerArr2 = []
+
+  // Checking for any number in input
+  function checkString(string) {
+    return /^[0-9]*$/.test(string)
```
aayushlama4008, line 12: I don't understand the use of   /^[0-9]*$/

src/scrabble.js
yee0802: Thanks :)

src/scrabble.js
yee0802: It's RegEx. Checking if there's any numbers (0-9) in the string given, on line 15.

PR #151, opened 2023-09-13
Daniel Roberts

irlydo:
Completed main exercise but not extension

### DIFF
--- a/src/scrabble.js
+++ b/src/scrabble.js
@@ -1,5 +1,48 @@
-function scrabble() {
-  // write code here
-}
+function scrabble(word) {
+  const newAlphabet = {
+    a: 1,
+    e: 1,
+    i: 1,
+    o: 1,
+    u: 1,
+    l: 1,
+    n: 1,
+    r: 1,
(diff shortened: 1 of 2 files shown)


src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+  let i
+
+  if (word === null) {
+    return 0
+  }
+  word = word.toLowerCase()
+  for (i = 0; i < word.length; i++) {
+    sum += newAlphabet[word[i]] || 0
+  }
+  if (word.includes('{' && '}')) {
```
irlydo, line 40: Ignore this if statement as it was just me testing a part of the extension

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+    w: 4,
+    y: 4,
+    k: 5,
+    j: 8,
+    x: 8,
+    q: 10,
+    z: 10
+  }
+  let sum = 0
+  let i
```
yee0802, line 31: You could define this within the for loop

PR #152, opened 2023-09-13
Rebecca Noy

extra-jalapenos:


src/constants.js
```diff
@@ -0,0 +1,52 @@
...
+  valid: true,
+}
+
+const resetCharFactor = () => {
+  charFactor.current = 1
+  charFactor.valid = true
+  delete charFactor.previous
+}
+
+const getValue = (char) => (!!char.match(/[^A-Za-z]/) ? 0 : charValue[char.toUpperCase()])
```
ilham-saleh: using double exclamation mark and a question mark is shorter version of if...else statement?

PCapid3v: I like this if statement.
Why does it have a double exclamation mark?
What this .match() method does?

src/constants.js
```diff
@@ -0,0 +1,52 @@
...
+
+const updateCharFactor = (newFactor) => {
+  if (charFactor.current * newFactor >= 1) {
+    charFactor.previous = charFactor.current
+    charFactor.current *= newFactor
+  } else {
+    charFactor.valid = false
+  }
+}
+
```
PCapid3v: It's a brilliant way to approach the bracket comparison. I might copy it :).
I don't think there is too much to refactor it's already another level. really impressed

src/constants.js
extra-jalapenos: Yes, that's shorthand: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator

Condition ? then : else
is the general idea

extra-jalapenos: the double question mark is to ensure that it gives back either a true or false, even when the result of char.match(/[^A-Za-z]/) is null / undefined, an array or whatever

spec/extensions/scrabble.spec.js
```diff
@@ -53,5 +53,9 @@ describe("Scrabble", () => {
     it('returns 0 for incorrect tokens', () => {
       expect(scrabble('|d|og')).toEqual(0)
     })
```
PCapid3v, line 55: Like it.
What if I start the word from a closing bracket? like }d}og
Will the program still run?

PCapid3v (commented):
I am impressed by this project, and well done!
Not much to review. She is clearly at an advanced level :)

src/constants.js
extra-jalapenos: - The double exlamation mark is to ensure that it gives back either a true or false, even when the result of char.match(/[^A-Za-z]/) is null / undefined, an array or whatever
- the .match() method checks whether something matches a pattern – in this case "does not contain the characters A-Za-z". It's a so called regular expression.

src/constants.js
extra-jalapenos: please do copy – programming rarely is about originality but more about collaboration! :)

spec/extensions/scrabble.spec.js
extra-jalapenos: It will! at the very start I check for "the word contains only valid characters" and then go from there. This is where d|o|g gets thrown out for the pipe characters. it'll also throw out }d}og for having an uneven number of braket pairs / trying to lower the factor below 1 (you start at 1, and } tries to get it to 1/2, which sets charFactor.valid to false (line 43)

src/constants.js
```diff
@@ -0,0 +1,52 @@
...
+const resetCharFactor = () => {
+  charFactor.current = 1
+  charFactor.valid = true
+  delete charFactor.previous
+}
+
+const getValue = (char) => (!!char.match(/[^A-Za-z]/) ? 0 : charValue[char.toUpperCase()])
+
+const charFactorChangeTrigger = (char) => {
+  if (!!char.match(/[\[\]\{\}]/)) {
```
Luca-Terrazzan: I recommend using regexes sparingly, in this case you could have skipped this by using the `default` statement in your switchcase or by `return`ing instead of `break`ing

extra-jalapenos: Thank you for the feedback, is this because it's not very intuitive to read with the escaping or for another reason?

Luca-Terrazzan: Readability is definitely a reason, but mostly I'd say it's just safer. It's very easy to make mistakes while writing regexes and they can also expose your application to security issues (like [redos](https://en.wikipedia.org/wiki/ReDoS))

PR #154, opened 2023-09-13
Lukas Dembicki

PeachyOmnivore:


src/scrabble.js
```diff
@@ -1,5 +1,81 @@
...
+
+    if (word[i] === '{' && word[i] === true) {
+      sum += pointValues[word[i]] * 2
+    }
+
+    if (word[i] === '{word}' && word === true) {
+      sum += pointValues[word[i]] * 2
+    }
+
+    if (word[i] === '{word}' && word === true) {
```
Faiza-tech: Nice work ! 
Can you please explain me a bit . what actually " word[ i ] " doing here

PeachyOmnivore: Ah this is just me experimenting. Its all not real code and does nothing. But theoretically i would look at the individual indexes of the word and see if it includes the parenthesis.

src/scrabble.js
```diff
@@ -1,5 +1,81 @@
...
+  if (
+    inputWord === '' ||
+    inputWord === null ||
+    inputWord === ' \t\n' ||
+    typeof inputWord !== 'string'
+  ) {
+    return 0
+  }
+  // ENSURING KEYS GET ACCESSED
+  const word = inputWord.toUpperCase()
```
Faiza-tech, line 41: Good to see , use of comments in code

src/scrabble.js
Faiza-tech: Thank you for explanation

src/scrabble.js
```diff
@@ -1,5 +1,81 @@
...
+  Z: 10
 }
 
+function scrabble(inputWord) {
+  // GUARD CLAUSE
+  if (
+    inputWord === '' ||
+    inputWord === null ||
+    inputWord === ' \t\n' ||
+    typeof inputWord !== 'string'
```
EwbiDev, line 36: BIIIG guard clause, me likey 😍

src/scrabble.js
EwbiDev: I agree with Faiza on the good comments! Makes figuring out what the purpose of the code is much quicker.

src/scrabble.js
```diff
@@ -1,5 +1,81 @@
...
+  return sum
+}
+
+// WHAT TO DO WITH THIS?
+function parenthesis1(word, index) {
+  for (let i = 0; i < word.length; i++)
+    if (word[i] === '[' || word[i] === '{') {
+      return true
+    }
+}
```
EwbiDev: Cheeky little function, you're along the right lines with wanting to check if the brackets happen. What if you were to save the boolean into a variable?

src/scrabble.js
```diff
@@ -1,5 +1,81 @@
...
+  // ADDING POINTS
+  for (let i = 0; i < word.length; i++) {
+    sum += pointValues[word[i]]
+
+    // DOUBLE AND TRIPLE WORDS
+    //  WRITE A SEPERATE FUNCTION FOR CHECKING FOR PARENTHESIS..
+    if (word === '[]' && word[i] === true) {
+      // JUST WRITING NONSENSE TO TRY AND PICTURE IT / SUEDO CODE NEXT?
+      sum += pointValues[word[i]] * 3
+    }
```
EwbiDev: I see what you're trying to achieve here, I believe you were trying to see if the word contains a `'['` and a `']'` right? Are there any string methods which may be able to achieve that?

EwbiDev (commented):
Easy code to follow, good job! You're definitely most of the way to answering the extensions you just need the last few pieces.

src/scrabble.js
PeachyOmnivore: Nice, I actually just changed this!

PR #156, opened 2023-09-13
Eduard

EwbiDev:


src/scrabble.js
```diff
@@ -21,18 +25,22 @@ const stringArrayToScore = (stringArray) => {
...
+  const result = LETTERSCORES[charUpper] * MULTIPLE
+  console.log('result :>> ', result);
+  if (result) {
+    return result
+  } else {
+    return 0
+  }
 }
 
 const scoreMultiplicatorizer = (character) => {
```
jdm1991, line 49: What a name for a const. Outstanding.

pseudocode.md
```diff
@@ -1,13 +1,66 @@
 # Pseudocode
```
extra-jalapenos, line 1: I saw you didn't have comments yet and I only had 2 people with a pr, so I thought I'd give you feedback -- not sure how much i'll truly find tho :D

src/scrabble.js
```diff
@@ -1,8 +1,12 @@
 const { LETTERSCORES, SCOREMULTIPLIER } = require('./constants')
```
jdm1991: I would be interested to hear how you decided to go about structuring this.

pseudocode.md
```diff
@@ -1,13 +1,66 @@
...
 
-#### Add the steps you need to take to implement the requirements for Scrabble Challenge
+Add the steps you need to take to implement the requirements for Scrabble Challenge
 
-```
-1.
+## 1. Making function `findWordScore`
```
extra-jalapenos, line 5: It's really cool that you outlined this, I flat out didn't do that but it's probably really good to think about it before starting

src/constants.js
```diff
@@ -0,0 +1,57 @@
...
+const SCORE_MULTIPLIER = {
+  '{': 2,
+  '}': 1 / 2,
+  '[': 3,
+  ']': 1 / 3
+}
+
+const VALID_SCORE = true
+
+module.exports = {
```
extra-jalapenos, line 50: I 100% stole this idea from you, because splitting it into a constants / calc file only makes sense 👍🏻

src/scrabble.js
```diff
@@ -1,5 +1,130 @@
...
+const scrabble = (word) => {
+  const stringArray = stringToArray(word)
+  const numArray = stringArrayToScore(stringArray)
+
+  if (validScoreChecker3000tron()) {
+    return tallyScore(numArray)
+  } else {
+    return 0
+  }
+}
```
yee0802, line 120: I like how you split it up into multiple functions !

src/constants.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+const SCORE_MULTIPLIER = {
+  '{': 2,
+  '}': 1 / 2,
+  '[': 3,
+  ']': 1 / 3
+}
+
+const VALID_SCORE = true
```
extra-jalapenos, line 48: I see we have a man of culture 😂 I'm sure your code wouldn't let me feed it "inval}id{word", very good and beyond what was outlined!

src/scrabble.js
```diff
@@ -1,5 +1,130 @@
...
+    } else {
+      return findLetterScore(val)
+    }
+  })
+  numArray = numArray.filter((val) => !(val === undefined))
+  return numArray
+}
+
+const multiplyCheck = (character) => {
+  edgeCasesIDisagreeWith(character)
```
extra-jalapenos, line 31: 😂 😂 😂 😂

src/scrabble.js
```diff
@@ -1,5 +1,130 @@
...
+  logMultiplierUsage(character)
+  if (SCORE_MULTIPLIER[character]) {
+    return true
+  } else {
+    return false
+  }
+}
+
+const findLetterScore = (character) => {
```
extra-jalapenos, line 40: most of the functions from here to the bottom could be shortened and still readable with the shorthand for if then else: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator

src/scrabble.js
```diff
@@ -1,5 +1,130 @@
...
+  if (INCORRECT_TOKENS.includes(character)) {
+    VALID_SCORE = false
+  }
+}
+
+const tallyScore = (numArray) => {
+  if (numArray.length === 0) {
+    return 0
+  } else {
+    return numArray.reduce((acc, val) => acc + val)
```
extra-jalapenos, line 75: see, I knew I'd learn something from this – I'm gonna look into what this does

spec/function.spec.js
```diff
@@ -0,0 +1,127 @@
...
+  stringToArray,
+  multiplyCheck,
+  findLetterScore,
+  scoreMultiplicatorizer,
+  tallyScore
+} = require('../src/scrabble')
+
+describe('stringToArray', () => {
+  beforeEach(() => {
+    resetConstants()
```
extra-jalapenos, line 12: I'm getting PTSD from this because it caused so many weird errors for me when running the tests until I figured out that it had to do with that – wish I would've seen this earlier / or your question thread because it was a pain

spec/function.spec.js
```diff
@@ -0,0 +1,127 @@
...
+
+  it("return false for '()'", () => {
+    expect(multiplyCheck('()')).toBeFalse()
+  })
+})
+
+xdescribe('findLetterScore', () => {
+  beforeEach(() => {
+    resetConstants()
+    scoreMultiplicatorizer() // TODO: wanted to test function with MULTIPLE but it always ends up = 1, can't figure where it goes :(
```
extra-jalapenos, line 73: Did you resolve this and just forgot the comment?

src/constants.js
```diff
@@ -0,0 +1,57 @@
+const LETTER_SCORES = {
```
extra-jalapenos, line 1: @Callumhayden99 did this very cool thing where he had the key value pairs in seperate lines ordered by value first, it's more readable imo

extra-jalapenos (commented):
💯 🥇 that's all I've got to say here

src/scrabble.js
```diff
@@ -1,5 +1,130 @@
-function scrabble() {
-  // write code here
+const { LETTER_SCORES, INCORRECT_TOKENS } = require('./constants')
```
tomlukewarden, line 1: code is really well structured

src/scrabble.js
```diff
@@ -1,5 +1,130 @@
...
+  }
+}
+
+const logMultiplierUsage = (character) => {
+  const usageKeys = Object.keys(MULTIPLIER_USAGE)
+  if (usageKeys.includes(character)) {
+    MULTIPLIER_USAGE[character]++
+  }
+}
+
```
tomlukewarden, line 64: good use of variables and making clear (and also silly hehe) names

tomlukewarden (commented):
really good way to solve this!

src/scrabble.js
extra-jalapenos: it's the validScoreChecker3000tron for me 😂

src/scrabble.js
EwbiDev: Thank you! 😋

src/scrabble.js
EwbiDev: Yeah of course! Happy to go over it later in a call or something. I started off with detailing what I wanted to do in my [pseudocode](https://github.com/Uprising5034/js-scrabble-challenge/blob/main/pseudocode.md) then it was a case of converting that pseudocode I planned into Javascript.

Essentially I break down the problem into small tasks which I then later made into functions, then identify what input they need and what they should be outputting.

src/scrabble.js
EwbiDev: Thanks! It didn't always look like that, was definitely more messier earlier on. 😅

pseudocode.md
EwbiDev: Thanks! Code writing is a lot easier if you don't have to worry about what the code is meant to be doing. I'd highly recommend trying it!

src/constants.js
EwbiDev: Haha, you're welcome 😉

src/constants.js
EwbiDev: No I never actually took into account backwards brackets actually 😱.
As the code currently is `'id'` would be calculated at 1/2 multiplier score and the rest will just have a multiplier value of 1.

src/constants.js
EwbiDev: I agree it's far more readable having the value as the key. It was something I considered doing but I ultimately decided just on a single key paired with a single value. Made working with the functions easier imo.

pseudocode.md
EwbiDev: I'm honored, you clearly found enough from the looks of things. 😄

spec/function.spec.js
EwbiDev: No, hasn't been resolved. The first test is only returning `1` rather than `2` as it should be doing. Putting it down to not knowing exactly how Jasmine is working really.

src/scrabble.js
EwbiDev: Thank you! 🙂

src/scrabble.js
EwbiDev: Agreed, I've seen it in use but not tried it myself yet. Something to look into for next time when dealing with just `true` / `false` conditions.

src/scrabble.js
EwbiDev: First thing I do when dealing with a data type is just remind myself what methods it has. `.reduce` was perfect for finding sums in this case.

spec/function.spec.js
EwbiDev: Definitely! I consistently had half the tests in extension fail which led me to create `resetConstants()` it wasn't till later I also added it to use in the spec files after some help from Jules and Lewis.

EwbiDev (commented):
Thanks for all the lovely comments 🙂

PR #158, opened 2023-09-13
Pierluigi

PCapid3v:


src/scrabble.js
```diff
@@ -1,5 +1,47 @@
...
+  Z: 10
+}
+function scrabble(word) {
+  let score = 0
+  if (word === null || word.length === 0 || word === '\n\t') {
+    console.log('checking invalid input')
+    return 0
+  }
+  // loop through the object
+  for (const key in letterValues) {
```
ilham-saleh: a good example of using for...in

PR #159, opened 2023-09-13
Tom Warden

tomlukewarden:


src/scrabble.js
```diff
@@ -1,5 +1,48 @@
-function scrabble() {
-  // write code here
+const scrabbleValue = {
```
Callumhayden99, line 1: I like the layout of the list since it's easy on the eye's and very easy to understand

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+  W: 4,
+  Y: 4,
+  K: 5,
+  J: 8,
+  X: 8,
+  Q: 10,
+  Z: 10
+}
+
+const scrabble = (makeWord) => {
```
Callumhayden99, line 30: I like how you kept it simple and didn't use load's of extra code when it's not needed. This made it very easy to understand, especially for a new coder like myself.

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+  if (
```
Callumhayden99, line 31: I liked how you used the if statement for this part, made it very clear and easy to understand

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+  ) {
+    return 0
+  }
+  const printWord = makeWord.toUpperCase()
+  let sum = 0
+
+  for (let i = 0; i < makeWord.length; i++) {
+    sum += scrabbleValue[printWord[i]]
+  }
+  return sum
```
Callumhayden99, line 45: Definitely not going to say anything to improve since i'm new to this but one thing i learnt myself is that you can use " if(!makeWord) return 0 " and this would of done your if statement for you i think?

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+const scrabble = (makeWord) => {
+  if (
+    makeWord === `    ` ||
+    makeWord === ` \t\n` ||
+    makeWord === null ||
+    typeof makeWord !== `string`
+  ) {
+    return 0
+  }
+  const printWord = makeWord.toUpperCase()
```
Satokii, line 39: I like your naming convention, using words to clearly state what everything is so you can keep track through your code!

src/scrabble.js
Satokii, line 1: Nice concise code throughout which is easy to understand

src/scrabble.js
tomlukewarden: i will give that a shot!

PR #160, opened 2023-09-13
Satoki

Satokii:


src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+    W: 4,
+    Y: 4,
+    K: 5,
+    J: 8,
+    X: 8,
+    Q: 10,
+    Z: 10
+  }
+
+  if (
```
extra-jalapenos, line 31: instead of excluding all known cases that should fail, you could make the criteria to only let the word pass if it contains allowed characters (A-Z, a-z, [, ], {, }) -- then you're definitely safe

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+    myWord === '' ||
+    myWord === ' \t\n' ||
+    myWord === null ||
+    myWord === undefined
+  ) {
+    return 0
+  } else {
+    let wordScore = 0
+
+    const newWord = myWord.toUpperCase().split('')
```
extra-jalapenos: I'd suggest naming the variable here something that indicates that this is an array – also maybe making this an anonymous function with const lettersArray = (word) => …?

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+
+    for (let i = 0; i < newWord.length; i++) {
+      const wordLetters = newWord[i]
+
+      if (alphabetPoints.hasOwnProperty(wordLetters)) {
+        const letterValues = alphabetPoints[wordLetters]
+        wordScore += letterValues
```
extra-jalapenos: Love the use of the shorthand 👍🏻  :)

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+  } else {
+    let wordScore = 0
+
+    const newWord = myWord.toUpperCase().split('')
+
+    for (let i = 0; i < newWord.length; i++) {
+      const wordLetters = newWord[i]
+
+      if (alphabetPoints.hasOwnProperty(wordLetters)) {
+        const letterValues = alphabetPoints[wordLetters]
```
extra-jalapenos: you don't really need to define a temporary variable here, just replace the "letterValues" in the next line with alphabetPoints[wordLetters] – it's still readable but shorter then

extra-jalapenos (commented):
Readable and understandable code :)

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+
+  if (
+    myWord === '' ||
+    myWord === ' \t\n' ||
+    myWord === null ||
+    myWord === undefined
+  ) {
+    return 0
+  } else {
+    let wordScore = 0
```
PCapid3v, line 39: Did exactly the same, maybe we are just lazy but surely there is  cooler way to do it :)

PCapid3v (commented):
I don't really know what to say it's quite similar to mine even though we find the way to make it works and keep it simple.
Great work!

src/scrabble.js
extra-jalapenos: What happens if you run scrabble("\t") though? or scrabble("\n")? will it still return 0?

Satokii: Yup still returns 0, I'll definitely try all your suggestions though thank you :)

PR #161, opened 2023-09-14
callum hayden

Callumhayden99:


src/scrabble.js
```diff
@@ -1,5 +1,35 @@
-function scrabble() {
-  // write code here
+function scrabble(word) {
```
Satokii, line 1: All code is very concise and easy to understand, nice work!

src/scrabble.js
```diff
@@ -1,5 +1,35 @@
...
+  if (!word) return 0;
+  let alphabet = {
+ 
+    'A': 1, 'E': 1, 'I': 1, 'O': 1, 'U': 1, 'L': 1, 'N': 1, 'R': 1, 'S': 1, 'T': 1,
```
extra-jalapenos, line 5: Really good idea to group the characters by value -- SO much better readable for humans! I'm 100% gonna nick that

src/scrabble.js
```diff
@@ -1,5 +1,35 @@
-function scrabble() {
-  // write code here
+function scrabble(word) {
+  if (!word) return 0;
```
extra-jalapenos, line 2: I'm not sure this is 100% for stopping weird input – for example, it would let weird characters pass or (probably) arrays, objects… I'd check for the word containing only allowed characters to make sure

src/scrabble.js
```diff
@@ -1,5 +1,35 @@
...
+
+  
+word = word.toUpperCase();
+let wordScore = 0;
+
+ for (let i = 0; i < word.length; i++) {
+ const letter = word[i];
+
+
+ if (alphabet[letter] !== undefined) {
```
extra-jalapenos, line 22: see comment in line 2, if you have weird characters but only catch them here, it'll still add the valid ones and give back a result (maybe test with "cäbbage" to see how you can improve line #2)

extra-jalapenos (commented):
Short and readable, only is very lax about what you can feed it as input

src/scrabble.js
```diff
@@ -1,5 +1,35 @@
...
+  let alphabet = {
+ 
+    'A': 1, 'E': 1, 'I': 1, 'O': 1, 'U': 1, 'L': 1, 'N': 1, 'R': 1, 'S': 1, 'T': 1,
+    'D': 2, 'G': 2,
+    'B': 3, 'C': 3, 'M': 3, 'P': 3,
+    'F': 4, 'H': 4, 'V': 4, 'W': 4, 'Y': 4,
+    'K': 5,
+    'J': 8, 'X': 8,
+    'Q': 10, 'Z': 10
+  };
```
PCapid3v, line 12: It's concise and clean. I like it!
Did Prettier suggest it, or did you?

src/scrabble.js
```diff
@@ -1,5 +1,35 @@
...
+word = word.toUpperCase();
+let wordScore = 0;
+
+ for (let i = 0; i < word.length; i++) {
+ const letter = word[i];
+
+
+ if (alphabet[letter] !== undefined) {
+  wordScore += alphabet[letter];
+ }
```
PCapid3v, line 24: Nice comparison!
Did you have any problem making it work with null as an input?

src/scrabble.js
```diff
@@ -1,5 +1,35 @@
...
+    'B': 3, 'C': 3, 'M': 3, 'P': 3,
+    'F': 4, 'H': 4, 'V': 4, 'W': 4, 'Y': 4,
+    'K': 5,
+    'J': 8, 'X': 8,
+    'Q': 10, 'Z': 10
+  };
+
+  
+word = word.toUpperCase();
+let wordScore = 0;
```
PCapid3v, line 16: I like the use of a name variable to clarify what score it is.

PCapid3v (commented):
Great job!

src/scrabble.js
extra-jalapenos: Ok so basically what I meant by this – the current logic is "if the opposite of the input is true, return 0". But as you have a very specific profile on what kind of input you want to allow, I'd take an approach that checks more selectively – it should throw out everything that is not a word containing nothing but allowed characters, which are: A-Z, a-z, and the different types of brakets: [ ] { }.

here's a hint on how you check for contents of a string when you don't know exactly what the string will be: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

extra-jalapenos: Regular expressions can be a lot at first but it's best explained at defining a pattern that you look for instead of exactly knowing what you want. If you want to see the described logic in code, have a look [at my file at line 11](https://github.com/boolean-uk/js-scrabble-challenge/pull/152/files#diff-df48340736af99ff21a33d27c2cdc9190750f76abe510659a40943f897880624)

PR #163, opened 2023-09-14
Jack Miller

jdm1991:


src/scrabble.js
```diff
@@ -1,5 +1,31 @@
...
+  
+  if (word === null) {
+    return 0
+  }
+  if (word === '' || word === " \t\n") {
+    return 0
+  }
+  word = word.toUpperCase();
+  const lettervalues = {
+  'A': 1, 'E': 1, 'I': 1, 'O': 1, 'U': 1, 'L': 1, 'N': 1, 'R': 1, 'S': 1, 'T': 1,
```
Callumhayden99, line 13: I like the layout of this since its really easy to understand and easy on the eye's

src/scrabble.js
```diff
@@ -1,5 +1,31 @@
-function scrabble() {
-  // write code here
+function scrabble(scrabbleword) {
```
Callumhayden99, line 1: overall you smashed it since you used minimal code to get the job done

src/scrabble.js
```diff
@@ -1,5 +1,31 @@
...
+
+  let word = scrabbleword
+  
+  if (word === null) {
```
Callumhayden99, line 5: im not 100% with what i'm about to say, but could you put the if statement as one or does it have to be two different one's?

src/scrabble.js
```diff
@@ -1,22 +1,30 @@
 function scrabble(scrabbleword) {
-  scrabbleword = scrabbleword.toUpperCase();
-if (scrabbleword === '' || scrabbleword === " \t\n" || scrabbleword === null) {
```
tomlukewarden: good way to start and finish the first checks

src/scrabble.js
Satokii: I would second Callum on this part, both `if` statements are returning 0 so you could potentially combine them into a single line of code using another `||` comparison

tomlukewarden (commented):
callum said most of what i was going to! good job and easy to understand code!

src/scrabble.js
```diff
@@ -1,5 +1,31 @@
...
+  'A': 1, 'E': 1, 'I': 1, 'O': 1, 'U': 1, 'L': 1, 'N': 1, 'R': 1, 'S': 1, 'T': 1,
+  'D': 2, 'G': 2,
+  'B': 3, 'C': 3, 'M': 3, 'P': 3,
+  'F': 4, 'H': 4, 'V': 4, 'W': 4, 'Y': 4,
+  'K': 5,
+  'J': 8, 'X': 8,
+  'Q': 10,'Z': 10,
+  }
+  let score = 0
+
```
Satokii, line 22: The grouping of different parts of your code that relate to each other so they are close together make it very easy to see what's going on which is great.

src/scrabble.js
```diff
@@ -1,5 +1,31 @@
...
+  'Q': 10,'Z': 10,
+  }
+  let score = 0
+  
+  for (let i = 0; i < word.length; i++)
+  {
+    score += lettervalues[word[i]]
+  }
+  return score
 }
```
Satokii, line 29: Nice concise and straightforward code overall, lovely mate

PR #164, opened 2023-09-14
Nazar Tymiv

NazarTymiv:


src/scrabble.js
```diff
@@ -1,5 +1,165 @@
-function scrabble() {
-  // write code here
+const scoreLibrary = [
```
julesnuggy, line 1: Cool name for the variable

src/scrabble.js
```diff
@@ -11,16 +11,153 @@ const scoreLibrary = [
...
       }
-    })
-  })
+
+      return scores
+    }
+
+    // CURLY BRACKETS
+    if (splitStr.includes('{') && splitStr.includes('}')) {
+      const curlyBrackets = str.toUpperCase().match(/{([^}]+)}/g)
```
PeachyOmnivore, line 106: hieroglyphics!!! Absolutely insane!

src/scrabble.js
```diff
@@ -1,5 +1,165 @@
...
+
+  let scores = 0
+
+  const splitStr = str.split('')
+
+  if (
+    (splitStr.includes('{') && splitStr.includes('}')) ||
+    (splitStr.includes('[') && splitStr.includes(']')) ||
+    /^[A-Z]*$/.test(str.toUpperCase())
+  ) {
```
julesnuggy, line 22: This could be extracted to a variable/function.

src/scrabble.js
```diff
@@ -1,5 +1,165 @@
...
+            item.split('').forEach((value) => {
+              scoreLibrary.forEach((scoreItem) => {
+                if (scoreItem.letters.includes(value)) {
+                  scores += scoreItem.score * 3
+                }
+              })
+            })
+          })
+
+        scores = scores * 2
```
julesnuggy, line 52: This could be `scores *= 2`

julesnuggy (commented):
Nice work!

src/scrabble.js
```diff
@@ -1,5 +1,165 @@
-function scrabble() {
-  // write code here
+const scoreLibrary = [
+  { letters: ['A', 'E', 'I', 'O', 'U', 'L', 'N', 'R', 'S', 'T'], score: 1 },
+  { letters: ['D', 'G'], score: 2 },
+  { letters: ['B', 'C', 'M', 'P'], score: 3 },
+  { letters: ['F', 'H', 'V', 'W', 'Y'], score: 4 },
+  { letters: ['K'], score: 5 },
+  { letters: ['J', 'X'], score: 8 },
+  { letters: ['Q', 'Z'], score: 10 }
```
ps975076, line 8: Nice way of assigning the scores to the letters. Tell me more 👀
I also used an object but like this:
"const letterScoring = {
A: 1
E: 1..."
Yours looks cleaner

src/scrabble.js
```diff
@@ -1,5 +1,165 @@
...
+  const splitStr = str.split('')
+
+  if (
+    (splitStr.includes('{') && splitStr.includes('}')) ||
+    (splitStr.includes('[') && splitStr.includes(']')) ||
+    /^[A-Z]*$/.test(str.toUpperCase())
+  ) {
+    let newStr = str
+
+    // MULTIPLE BRACKETS
```
ps975076, line 25: Loved how you're using comments to breakdown sections - really easy to read and navigate towards

src/scrabble.js
```diff
@@ -1,5 +1,165 @@
...
+  if (typeof str !== 'string') return 0
+
+  let scores = 0
+
+  const splitStr = str.split('')
+
+  if (
+    (splitStr.includes('{') && splitStr.includes('}')) ||
+    (splitStr.includes('[') && splitStr.includes(']')) ||
+    /^[A-Z]*$/.test(str.toUpperCase())
```
ps975076, line 21: Cool method used - what does the .test do?

ps975076 (commented):
Overall really great and clean work! :) 👍

src/scrabble.js
Chloe070196: Same question here :)  Does it do something like only applying .toUpperCase() to a string if necessary?

src/scrabble.js
NazarTymiv: Thanks :) It is really useful to do structure like me it is just array where I put objects inside each object there is key letters also array where I puted letters and another key it is score where I puted score for each letter from previous key "letters"
Is is good because when you make loop for search letter and score for this letter this structure will work faster and you will write less code for search function.

src/scrabble.js
NazarTymiv: Thanks :)

src/scrabble.js
```diff
@@ -1,5 +1,165 @@
...
+
+    // MULTIPLE BRACKETS
+    if (
+      splitStr.includes('{') &&
+      splitStr.includes('}') &&
+      splitStr.includes('[') &&
+      splitStr.includes(']')
+    ) {
+      let curlyBrackets = str.toUpperCase().match(/{([^}]+)}/g)[0]
+      let squareBrackets = str.toUpperCase().match(/\[(.*?)\]/g)[0]
```
yee0802, line 33: Nice use of RegEx in all your code!

src/scrabble.js
NazarTymiv: This method just check pattern which I used before calling this method. In my case this mathod just check if inside variable 'str' only characters if there is will be numbers of any another characters this test return false

src/scrabble.js
NazarTymiv: Thanks :)

src/scrabble.js
NazarTymiv: > Same question here :) Does it do something like only applying .toUpperCase() to a string if necessary?

Yes in this test I used just Capital letters and you should call .toUpperCase() before using method .test

src/scrabble.js
ps975076: Thank you for explaining that to me - this is really helpful

src/scrabble.js
```diff
@@ -1,5 +1,165 @@
...
+                  scores += scoreItem.score * 2
+                }
+              })
+            })
+          })
+
+        scores = scores * 3
+
+        squareBrackets
+          .slice(0, -1)
```
Luca-Terrazzan, line 87: You could have avoided a lot of copypasting by extracting this to a function

PR #165, opened 2023-09-14
aayush

aayushlama4008:


src/scrabble.js
```diff
@@ -1,5 +1,51 @@
-function scrabble() {
-  // write code here
+const scrabble = (myWord) => {
```
ps975076, line 1: Very clean code! Love it :)

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
...
+    myWord === '\t\n' ||
+    myWord === null ||
+    myWord === undefined
+  ) {
+    return 0
+  } else {
+    let wordpoint = 0
+    const splitedWords = myWord.toLowerCase().split('')
+    for (let i = 0; i < splitedWords.length; i++) {
+      const singleLetter = splitedWords[i]
```
ps975076, line 41: Like the use of the loop here.
Tell me more about the split method 👀

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
-function scrabble() {
-  // write code here
+const scrabble = (myWord) => {
+  const wordValues = {
```
ps975076, line 2: Could add some comments to outline each section

ps975076 (commented):
Overall, really nice clean and clear code! Good work! :)

src/scrabble.js
Chloe070196, line 1: I really like that you went with storing all the letter values as key-value pairs in an object - it's a lot more concise than my 7 arrays, and I can see how that also allows for the rest of your code to also be quite a bit more concise than what I went for.

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
...
+    y: 4,
+    k: 5,
+    j: 8,
+    x: 8,
+    q: 10,
+    z: 10
+  }
+  if (
+    myWord === '' ||
+    myWord === '\t\n' ||
```
Chloe070196, line 32: That one was tricky ('\t\n') - it is a string, and it has a length of two characters, yet it is just whitespace. Your line 32 seems like a good way to pass the related jasmine test! Still,  I'm curious to know whether your if statement here would also filter our longer ''empty'' strings.

src/scrabble.js
```diff
@@ -1,5 +1,51 @@
...
+    myWord === null ||
+    myWord === undefined
+  ) {
+    return 0
+  } else {
+    let wordpoint = 0
+    const splitedWords = myWord.toLowerCase().split('')
```
Chloe070196, line 39: I'd never have thought of using .split() here - nice!

PR #166, opened 2023-09-14
Faiza

Faiza-tech:


src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+    Y: 4,
+    K: 5,
+    J: 8,
+    X: 8,
+    Q: 10,
+    Z: 10
+  }
+  console.log(letterValue)
+  if (!word || typeof word !== 'string') {
+    return 0
```
EwbiDev, line 32: Nice stringing together multiple things into one guard clause.

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+  } else {
+    let score = 0
+    const letter = word.toUpperCase().split('')
```
EwbiDev, line 35: using `.split()' to make this into an array is a good idea! 🙂

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+    V: 4,
+    W: 4,
+    Y: 4,
+    K: 5,
+    J: 8,
+    X: 8,
+    Q: 10,
+    Z: 10
+  }
+  console.log(letterValue)
```
EwbiDev, line 30: `console.log()` is useful for when building your code for troubleshooting. But can make things a bit busy if you leave them around all the time. I would suggest getting rid of them once you're done with them.

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+  if (!word || typeof word !== 'string') {
+    return 0
+  } else {
+    let score = 0
+    const letter = word.toUpperCase().split('')
+    for (let i = 0; i < letter.length; i++) {
```
EwbiDev, line 36: Since `letter` is an array, you could have used an array method here instead such as `.forEach()`

EwbiDev (commented):
Nice job! 😇

src/scrabble.js
```diff
@@ -1,5 +1,48 @@
...
+      const singleLetter = letter[i]
+      if (letterValue[singleLetter]) {
+        const newScore = letterValue[singleLetter]
+        score = score + newScore
```
yee0802, line 40: This could be score += newScore !

src/scrabble.js
Faiza-tech: Yes you are right , but i used object in the start so array method or property made a problem here so for get rid of that i used this.

src/scrabble.js
Faiza-tech: yes i could be

PR #167, opened 2023-09-14
Tom Eastwood 

TomEastwood:


src/scrabble.js
```diff
@@ -1,5 +1,105 @@
...
+        initialScore = 0
+        break
+      }
+    } else {
+      initialScore = 0
+      break
+    }
+  }
+  return initialScore
+}
```
NazarTymiv, line 83: You can use here word.split('') it will make the array with elements inside of each letter from word and after that you can use forEach method for take every letter from this array :)

spec/extensions/scrabble.spec.js
NazarTymiv, line 1: I like it. It is very useful efull for debugging. Good job :)

NazarTymiv (commented):
very beautiful and clear code :)

src/scrabble.js
```diff
@@ -1,5 +1,105 @@
...
+
+function Checker2(word, idx) {
+  let check = false
+  for (let i = idx; i >= 0; i--) {
+    if (word[i] === '{' || word[i] === '[') {
+      check = true
+    }
+  }
+  return check
```
PeachyOmnivore, line 102: I like these. Its the approach I am currently dabbling in.

src/scrabble.js
```diff
@@ -1,5 +1,105 @@
...
+    const letter = word.charAt(idx).toUpperCase()
+
+    if (letterValues[letter] !== undefined) {
+      initialScore += letterValues[letter] * multiplier
+    } else if (letter === '{') {
+      if (Checker1(word, idx) === true) {
+        multiplier = multiplier * 2
+      } else {
+        initialScore = 0
+        break
```
PeachyOmnivore, line 54: I would like to know more about these `breaks` here. This is just to prevent the multiplier from continuing to multiply?

src/scrabble.js
```diff
@@ -1,5 +1,105 @@
...
+      }
+    } else {
+      initialScore = 0
+      break
+    }
+  }
+  return initialScore
```
PeachyOmnivore, line 82: I wonder if all this could be another function in another files and imported over. So all you have to do is call that fucntion from within your scrabble function?

PeachyOmnivore (commented):
Amazing work. I wish I got this far!

PR #168, opened 2023-09-15
Priya

ps975076:


src/scrabble.js
```diff
@@ -1,5 +1,131 @@
...
+console.log('-----------------------------------------------')
+const newLine = 'hello\nWorld!'
+console.log('newLine:', newLine)
+// newLine: hello
+// World!
+
+const tab = 'Hello\tWorld'
+console.log("tab:", tab)
+// tab: Hello      World
+
```
NazarTymiv, line 129: Really beautiful and useful debugging good job :)

src/scrabble.js
```diff
@@ -1,5 +1,131 @@
...
+  Z: 10
 }
 
+
+// 2. Create a function called 'scrabble' with one parameter called 'word'
+function scrabble(word) {
+
+
+
+  if (word === null || word === ' \t\n' || word === '') {
```
NazarTymiv, line 82: On my mind it will be better to use just typeof and check if the variable word is string and delete word === null, word === ''

src/scrabble.js
```diff
@@ -1,5 +1,131 @@
...
+// If statements ensuring if:
+// - '' is entered should return 0
+// - \t\n is entered should return 0
+// - null should return 0
+
+// 5. Create a variable that keeps track of the score
+
+// 6. Create a loop to loop through each word entered
+// - if any word entered
+// add the totalScore to the scoreLetters
```
NazarTymiv, line 43: Ohh that's very good to do plan for your exercise it is very big and clear plan good job I like it :)

NazarTymiv (commented):
Greate work and relly good start :)

src/scrabble.js
ps975076: Ok, do you mean check if the word is an empty string " ' ' ", rather than typing the word 'null'?

src/scrabble.js
ps975076: Thanks so much

PR #178, opened 2023-09-17
Hamza Ahmed

Hamza789987:


src/scrabble.js
```diff
@@ -1,5 +1,98 @@
...
+
+function scrabble(input) {
+  let counter = 0
+  if (input === null || input.length === 0) {
+    return counter
+  }
+  for (let i = 0; i < input.length; i++) {
+    const eachLetter = input[i]
+    if (
+      eachLetter === `a` ||
```
Luca-Terrazzan, line 14: Try to use [.toLowerCase()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase) to halve your checks, as well as having a data structure to contain your letter scores. This way you should be able to reduce your lines of code as well as reducing your total copy-pasting

PR #183, opened 2023-09-18
Atiq07

Atiq07:


src/scrabble.js
```diff
@@ -1,5 +1,77 @@
...
+    Y: 4,
+    K: 5,
+    J: 8,
+    X: 8,
+    Q: 10,
+    Z: 10
+  }
+  
+  // 2. Create a function called 'scrabble' with one parameter called 'word'
+  function scrabble(word) {
```
Luca-Terrazzan: You are copying the scrabble function within another scrabble function. This means you are not returning anything when `scrabble()` is invoked from outside this file. This is why all your tests are failing. Please update so that you have only one `scrabble()` function and only one module export

PR #198, opened 2024-02-27
Rafa Lopes

rafa-lopes-pt:


src/Logic/player.ts
```diff
@@ -0,0 +1,125 @@
...
+  rack: Rack
+
+  constructor(id: string) {
+    this._id = id
+    this.name = 'John Doe'
+    this.randomizeColor()
+    this.randomizeAvatar()
+    this.points = 0
+    //time and rack are initialized only by the game
+    this.rack = new Rack()
```
vherus: Hard defined dependency. Move to dependency injection

```js
constructor(id: string, rack: Rack = new Rack()) {
```

src/Logic/player.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  shuffle() {
+    for (
+      let currentIndex = this._rack.length - 1;
+      currentIndex > 0;
+      currentIndex--
+    ) {
+      let randIndex = Math.floor(Math.random() * (currentIndex + 1))
+      var temp = this._rack[currentIndex]
```
vherus: Naughty

PR #210, opened 2024-03-01
Shaun Harris

Shaun-Harris:
Currently running with 1 error submitting for the morning task. current failure:-

".......F

Failures:
1) Scrabble scores a case-insensitive word
  Message:
    Expected 0 to equal 41."

### DIFF
--- a/src/scrabble.js
+++ b/src/scrabble.js
@@ -27,33 +27,28 @@ const lettersValues = {
 }
 
 const scrabble = (word) => {
-  if (wordValidate(word)) {
-    let lowerCaseWord = word.toLowerCase()
-    let wordScoreTally = tallyLetterScores(lowerCaseWord)
-    let wordScoreDouble = doubleWordRule(lowerCaseWord)
-    let wordScoreTriple = tripleWordRule(lowerCaseWord)
-    let totalScore = tallyFinalScores(
-      wordScoreTally,
-      wordScoreDouble,
-      wordScoreTriple
-    )
-    return totalScore
+  if (validateWord(word)) {
+    const lowerCaseWord = word.toLowerCase()
+    const wordScoreTally = tallyLetterScores(lowerCaseWord)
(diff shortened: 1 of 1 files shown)


src/scrabble.js
```diff
@@ -1,5 +1,146 @@
...
+  if (wordValidate(word)) {
+    let lowerCaseWord = word.toLowerCase()
+    let wordScoreTally = tallyLetterScores(lowerCaseWord)
+    let wordScoreDouble = doubleWordRule(lowerCaseWord)
+    let wordScoreTriple = tripleWordRule(lowerCaseWord)
+    let totalScore = tallyFinalScores(
+      wordScoreTally,
+      wordScoreDouble,
+      wordScoreTriple
+    )
```
vherus:
```js
return tallyFinalScores(
      wordScoreTally,
      wordScoreDouble,
      wordScoreTriple
    )
```

src/scrabble.js
```diff
@@ -1,5 +1,146 @@
...
+  }
+}
+
+const wordValidate = (word) => {
+  let openingBrace = 0
+  let closingBrace = 0
+  let openingBracket = 0
+  let closingBracket = 0
+
+  if (word === null) {
```
vherus:
```js
if (!word)
```

src/scrabble.js
```diff
@@ -1,5 +1,146 @@
...
+    if (
+      word[i] === '{' ||
+      word[i] === '}' ||
+      word[i] === '[' ||
+      word[i] === ']'
+    ) {
+      continue
+    }
+    if (word[i - 1] === '{' && word[i + 1] === '}') {
+      wordScoreDouble += (lettersValues[word[i]] * 2)
```
vherus: * 2

2 is a magic number

vherus: Move this to a variable so we know what it is.

PR #218, opened 2024-05-01
Thomas Jensen

th0jensen:


src/scrabble.js
```diff
@@ -1,5 +1,103 @@
-function scrabble() {
-  // write code here
+const modifiers = ['{', '}', '[', ']']
+let totalPoints
+const points = {
```
dogezen: when we have constants: variables storing contant data that is fixed:

```js
const SCRABBLE_POINTS = { ... }
```

src/scrabble.js
```diff
@@ -1,5 +1,103 @@
...
+  w: 4,
+  y: 4,
+  k: 5,
+  j: 8,
+  x: 8,
+  q: 10,
+  z: 10
+}
+
+function scrabble(input) {
```
dogezen: input doesn't tell me what the variable is: it would be better named as `inputWord`

src/scrabble.js
```diff
@@ -1,5 +1,103 @@
...
+  totalPoints = 0
```
dogezen, line 37: why is totalPoints defined outside the function and not created inside the function?

src/scrabble.js
```diff
@@ -1,5 +1,103 @@
-function scrabble() {
-  // write code here
+const modifiers = ['{', '}', '[', ']']
```
dogezen: same as scrabble points: i'd name these as `WORD_MODIFIERS`

src/scrabble.js
```diff
@@ -1,5 +1,103 @@
...
+
+function scrabble(input) {
+  totalPoints = 0
+
+  if (typeof input !== 'string') {
+    return totalPoints
+  }
+
+  if (modifiers.some((mod) => input.includes(mod))) {
+    square(input)
```
dogezen: unsure what square() and bracket() do -> function names are ambiguous, they should be more descriptive about telling me what the function does

src/scrabble.js
dogezen: function name ambiguous, better: `computeScrabbleScore`

dogezen (commented):
all your code is missing comments - it makes it hard for me to review because I have to analyise each line of code and really think hard about what each line does

with comments you make other people's lives much easier in terms of understanding and reflecting on your code

PR #222, opened 2024-05-02
core of challenge

NadProgress:


src/scrabble.js
```diff
@@ -1,5 +1,55 @@
-function scrabble() {
+function scrabble(words) {
```
dogezen, line 1: `words` is not correct, the variable represents 1 scrabble `word` so the name for the variable is better as `word`

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
   // write code here
-}
+  const dictionary = {
```
dogezen, line 3: the name of this variable doesnt tell me what it represents / stores, a better name would be `SCRABBLE_LETTER_SCORES` since it is representing a fixed-value piece of data that you only read from. You can move it outside the function since it is constant and read-only.

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+    S: 1,
+    T: 1,
+    U: 1,
+    V: 4,
+    W: 4,
+    X: 8,
+    Y: 4,
+    Z: 10
+  }
+  let fullTotal = 0
```
dogezen, line 31: `fullTotal` variable name could improve because it doesnt tell me what it is, i don't know what full total means... perhaps `wordTotalScore` would be better or `totalScore`

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+  if (typeof words !== 'string') {
+    return 0
+  }
+
+  if (words.length === 0) {
+    return 0
+  }
+  
+  if (words === null) {
```
dogezen, line 40: this if statement should be the first of the 3 that check whether the word is valid

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+    return 0
+  }
 
+  for (let index = 0; index < words.length; index++) {
+    let wordsCapit = words[index].toUpperCase()
```
dogezen, line 45: this variable is not correctly named: it isn't a word capitalised, it's the character capitalised...

PR #223, opened 2024-05-02
Atakan Temizkan

AtikoSpeed:


src/scrabble.js
```diff
@@ -1,5 +1,150 @@
-function scrabble() {
-  // write code here
+function checkIncorrect(letterArray) {
+  const allowedChars = [
```
dogezen, line 2: better put this outside function since its const and read only:

`SCABBLE_LETTERS` or similar would be better name

src/scrabble.js
```diff
@@ -1,5 +1,150 @@
-function scrabble() {
-  // write code here
+function checkIncorrect(letterArray) {
```
dogezen, line 1: checkIncorrect is bad name doesn't tell me what it is going to check for / what it is goign to return, pick a better name

dogezen, line 1: this is usually called character array not letter array

src/scrabble.js
```diff
@@ -1,5 +1,150 @@
...
+  if (!letterArray.includes('[') && letterArray.includes(']')) {
+    return 0
+  }
+
+  if (!letterArray.includes('{') && letterArray.includes('}')) {
+    return 0
+  }
+}
+
+function specialCases(letterArray) {
```
dogezen, line 56: again, bad function name, special cases doesn't tell me what this function is meant to do

src/scrabble.js
```diff
@@ -1,5 +1,150 @@
...
+    return 0
+  }
+
+  if (specialCases(letterArray) !== undefined) {
+    return specialCases(letterArray)
+  }
+
+  for (let i = 0; i < letterArray.length; i++) {
+    let currentLetterScore = 0
+    if (onePointers.includes(letterArray[i])) {
```
dogezen, line 110: this is quite a lot of code repetition, if you had represented the letters and scores as a dictionary, you could have done this in 3 lines of code:

```js
const LETTER_SCORES = {
  a: 1,
  b: 2,
  etc..
}

// score:
const score = LETTER_SCORES[letter]
```

PR #224, opened 2024-05-02
Olaoluwa Oyebode_Pull Request

olaolumcpaul:


src/scrabble.js
```diff
@@ -1,5 +1,55 @@
-function scrabble() {
-  // write code here
+function scrabble(word) {
+  if (!word || typeof word !== 'string' || word.trim() === '') return 0
+
+  const scores = {
```
dogezen, line 22: this can go outside the function; since it is constant variable that is read-only..

you can also rename it as `SCRABBLE_SCORES` or `SCRABBLE_LETTER_SCORES` to indicate that it is read only and be more specific about what it represents

src/scrabble.js
```diff
@@ -1,5 +1,55 @@
...
+    J: 8,
+    X: 8,
+    Q: 10,
+    Z: 10
+  }
+
+  let totalScore = 0
+  const upperCaseWord = word.toUpperCase()
+
+  for (let i = 0; i < upperCaseWord.length; i++) {
```
dogezen: ok

src/scrabble.js
olaolumcpaul: all noted.

